PR #1, opened 2017-03-03
Snapcraft package

cholcombe973:
This is the beginning of a snapcraft package for gluster-block.  It is currently missing targetcli and tcmu-runner.  This should allow gluster-block to deployed on many more platforms.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ DISTCLEANFILES = Makefile.in gluster-block.spec autom4te.cache
 CLEANFILES = *~ gluster-block.spec
 
 EXTRA_DIST = autogen.sh README.md COPYING-GPLV2 COPYING-LGPLV3 \
-						 gluster-block.spec INSTALL
+						 gluster-block.spec INSTALL NEWS
 
 if USE_SYSTEMD
 RPM_WITH_FLAGS = --with systemd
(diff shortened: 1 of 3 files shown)


nixpanic:
Hi @cholcombe973,

What is the use when targetcli and tcmu-runner are not available? I'm not familiar with Snapcraft, so I might be missing some basic understanding.

Thanks,
Niels

cholcombe973:
That's a really good question.  I think the answer is it's useless without
targetcli and tcmu-runner.  I started working on amending my PR but I never
got around to finishing it.  Snapcraft is basically an easy way to package
your software so you can deploy it across many different linux platforms.
Looks like centos is in progress: https://snapcraft.io/docs/core/install

nixpanic:
Thanks for the answer! I think we can hold off with merging this until
tcmu-runner and targetcli are available too. Or, maybe add some TODO
comments in the file so that it is clear what still is pending.

Also, you probably should use `configure.ac` to create this from a `.in`
file, that way the version can be automatically replaced (see the
`.spec.in` for an example of `@PACKAGE_VERSION@`).

I'm not sure if @pkalever accepts GitHub pull requests, some of the
other patches have been reviewed+merged through Gerrit at
https://review.gluster.org/

pkalever:
This seems quite old now, feel free to open a new PR based on master. Thanks!

PR #21, opened 2017-05-12
Make USER field available in Block Info.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -2547,6 +2547,7 @@ blockInfoCliFormatResponse(blockInfoCli *blk, int errCode,
     json_object_object_add(json_obj, "GBID", GB_JSON_OBJ_TO_STR(info->gbid));
     json_object_object_add(json_obj, "SIZE", json_object_new_int64(info->size));
     json_object_object_add(json_obj, "HA", json_object_new_int(info->mpath));
+    json_object_object_add(json_obj, "USER", GB_JSON_OBJ_TO_STR(info->userid));
     json_object_object_add(json_obj, "PASSWORD", GB_JSON_OBJ_TO_STR(info->passwd));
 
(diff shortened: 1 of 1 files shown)


pkalever:
It was an intentional :-)
The GBID is username in "gluster-block info" response.
so we don't want to duplicate GBID and USERNAME. 

Thanks for the effort though.

humblec:
@pkalever Auth enable reflect a new `USERNAME` field, why its so then? Also, how come one can assume "GBID" is the `USERNAME` from just looking at the `info` response?

pkalever:
I personally don't like to see duplicated fields in the info.
May be we can do below,
1. document GBID uses filename, username, wwid, iqn postfix ...
2. when auth is enable s/GBID/USERNAME/
3. In the worst case, we can print like "GBID/USERNAME: bla bla bla bla"

the order is w.r.t my choice, point 1 is most preferable.

@vbellur @pranithk Thoughts ?

humblec:
@pkalever I can adjust the patch to `2 or 3`. I think making `USERNAME` visible is valid in the outputs. Will wait for others opinion as well, before sending next version.

humblec:
@vbellur any thoughts ?

pranithk:
3rd option looks good to me.

vbellur:
3rd option looks good to me too, thanks!

pkalever:
@humblec  Please make the changes to reflect 3rd option and submit to our gerrit repository, as we are not ready for github yet.

The Current repository that we are using is gerrit, merging your changes with github will reflect into origin and these are all together different repos.

pkalever:
@humblec do you have plans to work on this soon ?

pkalever:
Please reopen this, if needed.

PR #69, opened 2018-04-05
saveconfig: only alter given block configuration

pkalever:
We use saveconfig in all the remote calls i.e. create/delete/modify/replace.

Currently, if we try to operate on a given block, the consequent saveconfig
will result in rewriting the whole node configuration again as per the current
state of targets in the machine, this could result in override of currently
existing but inactive blocks configuration.

This patch support saving configuration per block. i.e. only the configuration
of the block that is requested is altered, rest all will be untouched.

Thus, this patch can avoid overriding of the currently unloaded configurations.

### DIFF
--- a/daemon/gluster-blockd.c
+++ b/daemon/gluster-blockd.c
@@ -25,7 +25,7 @@
 # define   GB_TGCLI_GLOBALS     "targetcli set "                               \
                                 "global auto_add_default_portal=false "        \
                                 "auto_enable_tpgt=false loglevel_file=info "   \
-                                "logfile=%s && targetcli / saveconfig"
+                                "logfile=%s"
 
 
 extern size_t glfsLruCount;
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -30,7 +30,7 @@
 # define   GB_TGCLI_GLFS        "targetcli " GB_TGCLI_GLFS_PATH
 # define   GB_TGCLI_CHECK       GB_TGCLI_GLFS " ls | grep ' %s ' | grep '/%s ' > " DEVNULLPATH
 # define   GB_TGCLI_ISCSI_PATH  "/iscsi"
-# define   GB_TGCLI_SAVE        "/ saveconfig"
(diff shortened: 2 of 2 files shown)


pkalever:
review-request: @pranithk @lxbsz @vbellur

pranithk:
After discussion with @pkalever we agreed that we will add a dependency with the right upstream version of rtslib/configshell/targetcli to make sure this doesn't cause any problem. https://github.com/gluster/gluster-block/issues/76 is raised to track this for the next release.

pkalever:
@pranithk @lxbsz can one of you test the patch and confirm ?

lxbsz:
Comparing by using targetcli cmd directly and gluster-block tool, the new patches works well as expect for me.
Tested-by: xiubli@redhat.com

pkalever:
For merging this PR, I did

```
$ git fetch origin pull/69/head:pkalever-saveconfig
$ git checkout pkalever-saveconfig

# then re-based PR on top of origin master and 
# appended the commit msg with 'Tested-by:' and 'Reviewed-by:' signatures

$ git checkout master
$ git merge --no-ff pkalever-saveconfig
$ git push origin master
```

Pretty sure  I missed something but nit sure what :-(
I do see the commit/patch got merged, but the PR branch still shows open ?

pkalever:
The reason we want to use cli way of merging PR's is that we want to maintain a track of who Tested the PR, who reviewed and who Acked the PR.

@vbellur @nigelbabu @nixpanic @obnoxxx @raghavendra-talur @kshlm how do we fix this problem ? Is there a better practice ?

nixpanic:
> The reason we want to use cli way of merging PR's is that we want to
> maintain a track of who Tested the PR, who reviewed and who Acked the
> PR.

That's what Gerrit and other tools are for... Maybe you want to try
GerritHub.io because that integrates with GitHub PRs?

Once you're ammending commit messages, just add "Closes: #69" to that
list as well. When you push the commits, the PR will automatically get
closed. PRs and Issues use the same numbering, so a PR is an Issue, and
the other way around.

pkalever:
Merged already via CLI. Closing it

PR #77, opened 2018-04-11
rpc: glfs-operations: update api call for ftruncate

bhumikagoyal:
The updated gfapi for glusterfs fop ftruncate has pre/post attributes in the function signature. But this api call inside gluster-block is not updated, so update the api call for ftruncate to resolve build errors.

Signed-off-by: Bhumika Goyal <bgoyal@redhat.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,8 +80,10 @@ AC_CHECK_HEADERS([stdio.h stdlib.h string.h stdbool.h stddef.h \
 
 # Checks for libraries.
 # glusterfs-api versions have a prefix of "7."
-PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],,
-                  [AC_MSG_ERROR([gfapi library >= 3.6 is required to build gluster-block])])
+
+PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],
+                  [PKG_CHECK_MODULES([gfapi], [glusterfs-api < 7.4.1],[AC_DEFINE([USE_LEGACY_FTRUNCATE],1,[Define 1 if glusterfs-api version < 7.4.1])],true)],
+                 [AC_MSG_ERROR([gfapi library >= 3.6 is required to build gluster-block])])
 AC_SUBST(GFAPI_CFLAGS)
 AC_SUBST(GFAPI_LIBS)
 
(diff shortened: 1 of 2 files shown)


rpc/glfs-operations.c
```diff
@@ -174,7 +174,7 @@ glusterBlockCreateEntry(struct glfs *glfs, blockCreateCli *blk, char *gbid,
     ret = -1;
     goto out;
   } else {
-    ret = glfs_ftruncate(tgfd, blk->size);
+    ret = glfs_ftruncate(tgfd, blk->size, NULL, NULL);
```
pkalever: Should we add version check for not breaking with older versions
#if CONFIG_GLUSTERFS_VERSION < X.Y.Z
 ret = glfs_ftruncate(tgfd, blk->size);
#else
ret = glfs_ftruncate(tgfd, blk->size, NULL, NULL);
#end

bhumikagoyal: Yeah right. My change will break the build with older versions. Will put a check and update the PR.

rpc/glfs-operations.c
```diff
@@ -8,7 +8,7 @@
   cases as published by the Free Software Foundation.
 */
 
-
+# include <../config.h>
```
amarts: this doesn't look good! should be #include "config.h" or rather we should add '-include config.h' in CFLAGS in configure itself!

configure.ac
```diff
@@ -80,8 +80,10 @@ AC_CHECK_HEADERS([stdio.h stdlib.h string.h stdbool.h stddef.h \
 
 # Checks for libraries.
 # glusterfs-api versions have a prefix of "7."
-PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],,
-                  [AC_MSG_ERROR([gfapi library >= 3.6 is required to build gluster-block])])
+
+PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],
+                  [PKG_CHECK_MODULES([gfapi], [glusterfs-api < 7.4.1],[AC_DEFINE([USE_LEGACY_FTRUNCATE],1,[Define 1 if glusterfs-api version < 7.4.1])],true)],
```
amarts: with release of glusterfs-5.0 these numbers may change... please have a look!

pkalever:
As discussed at https://github.com/gluster/gluster-block/pull/150 we should close it.

pkalever:
Fixed by: https://github.com/gluster/gluster-block/pull/178

PR #81, opened 2018-04-24
config-backup: increase max backup files count

pkalever:
$ targetcli get global max_backup_files
max_backup_files=10000

### DIFF
--- a/daemon/gluster-blockd.c
+++ b/daemon/gluster-blockd.c
@@ -31,7 +31,8 @@
 # define   GB_TGCLI_GLOBALS     "targetcli set "                               \
                                 "global auto_add_default_portal=false "        \
                                 "auto_enable_tpgt=false loglevel_file=info "   \
-                                "logfile=%s auto_save_on_exit=false"
+                                "logfile=%s auto_save_on_exit=false "          \
+                                "max_backup_files=100"
 
 # define   VERNUM_BUFLEN        8
 


pkalever:
@pranithk @lxbsz 
do we need this to be configurable ? IMO 10000 is a big number already, we might not want to change this ?

lxbsz:
@pkalever @pranithk 

IMO, 10000 is big enough. BTW, the 10000 is a very large number, do we really need to increase it to 10000 ?

Sometime when we have hundreds of target objects, each back file will be >1M, so the backup files will take around 10G disk space.

pkalever:
@lxbsz till now I have seen max 15 devices per node, for the use cases that block is supported now.
Based on that info, I thought 10000 would be okay.

May be we want to cross check size that saveconfig.json consumes for 1024 block devices and then decide this ?

Thanks!

lxbsz:
@pkalever Yeah, agree.

From my env, with 3 target objects the backup file size is up to 20K. So I just want to know whether that will happen.

pkalever:
@lxbsz can you please try to find out the size of saveconfig.json for 100 blocks in HA 3 setup, and guide us here with some number ?

Thanks!

lxbsz:
@pkalever 

The following is my test result:

HA 2:

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
50
# du -sh /etc/target
12M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
100
# du -sh /etc/target
58M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
150
# du -sh /etc/target
135M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
500
# du -sh /etc/target
992M	/etc/target

HA 3:
# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
50
# du -sh /etc/target
19M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
100
# du -sh /etc/target
91M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
150
# du -sh /etc/target
202M	/etc/target

# targetcli ls|grep iqn.2016-12.org.gluster-block|wc -l
500
# du -sh /etc/target
1481M	/etc/target

From the result we can see that, it is nonlinear growth, if the count is 1000, for HA = 3 the disk size should be about a very large size, around 4G ~ 5G. 

So the max_backup_files shouldn't too large. IMO <=500 should be acceptable.

Thanks,

pkalever:
self note: remove the auto save on exit in this patch, while refresh.

pkalever:
https://github.com/open-iscsi/targetcli-fb/pull/124 should unblock this PR

amarts:
@pkalever lets get this in as the blocking PR is now merged.

pkalever:
@amarts sure, will check if the latest targetcli fedora package holds the fix. If it has the fix, then we can take this in.

Will get back soon on this.

Thanks!

pkalever:
@amarts the latest release version 2.1.fb49 doesn't have the targetcli fix:

```
commit 3d9e6c616ca7789d281843caf2d3dfb99dbf78a0 (refs/remotes/origin/compress, refs/heads/compress)
Date:   Wed Nov 28 16:55:38 2018 +0530

    saveconfig: compress the backup config files
    
    We have noticed saveconfig.json with 100 storage objects and 100 targets
    (each holding a single tpg and a portal) consumes disk space around ~500K.
    
    Which is very expensive, and backing-up such 100 saveconfig.json files will
    take ~50M of disk space under /etc/target/backup/
    
    And at scale like 1000 storage objects and targets, this will become worst.
    
    Hence this patch attempts to compress(gzip) and store saveconfig.json while
    backing-up.
    
    Saved space example:
    
    [root@localhost ~]# targetcli ls | grep -e "user:glfs" -e "iscsi"
      | o- user:glfs ......................... [Storage Objects: 100]
      o- iscsi ............................... [Targets: 100]
    
    [root@localhost ~]# du -sh /etc/target/saveconfig.json
    448K    /etc/target/saveconfig.json
    
    [root@localhost ~]# du -sh /etc/target/backup/saveconfig-20181128-18\:20\:43-json.gz
    12K     /etc/target/backup/saveconfig-20181128-18:20:43-json.gz
    
    Reducing disk usage per backup file from 448K to 12K is very efficient right.
    
```

should we still be waiting for a release which pull this fix in ?

Thanks!

pkalever:
@lxbsz @amarts Please review.

updated the max_backup_files to 100.
Hope we are good with it.

pkalever:
targetcli release targetcli-2.1.fb49 holds compressing backupfiles feature. Since gluster-block now checks for minimum targetcli-2.1.fb49 as dependency, we are good to push this patch now.

pkalever:
@lxbsz thanks for the review. Updated the tags, merging this now.

PR #85, opened 2018-05-11
[WIP] gluster-block: load-balance support

pkalever:
Problem:
-------
If we have say 4 blocks with HA 3, there is possibility that for all the
4 devices when logged at the initiator, active IO paths connect to only one
node out of 3, this is because we do not decide the priority of the target
portals for a device on the target side and set priority to constant
(all paths prio 1) at the initiator side.

So there is no guarantee of load balance among 3 nodes.

Solution:
-------
By setting one portal to active optimised (AO) alua access state and other to
active non-optimised (ANO), the priority of portal which is assigned to AO is
set to 50 and for ANO it is 10. So portals assigned to AO will be considered
to be active paths.

We maintain the data about the priority paths in each glusterfs volume inside
metadata directory

While we create a block, counter for respective prio portal is incremented

Ex:
$ for i in {1..4}; do \
  gluster-block create sample/block${i} ha 3 \
   192.168.124.149,192.168.124.162,192.168.124.184 1GiB; \
done

$ cat /mnt/block-meta/prio.info
192.168.124.149: 1
192.168.124.162: 1
192.168.124.184: 1
192.168.124.149: 2

While we delete a block, counter for respective prio portal is decremented

$ for i in {1..4}; do gluster-block delete sample/block${i}; done

$ cat /mnt/block-meta/prio.info
192.168.124.149: 1
192.168.124.162: 1
192.168.124.184: 1
192.168.124.149: 2
192.168.124.149: 1
192.168.124.162: 0
192.168.124.184: 0
192.168.124.149: 0


pkalever:
TODO:
Add capabilities.

pkalever:
@lxbsz @pranithk please review.

lxbsz:
I am thinking how about just update the old exist line to record the numbers of each node ? Though all the meta data recorded like currently we do, it should be more readable and faster to check this. If we have thousands of target devices, this file should have many lines to parse every time when creating and deleting.

Not very sure, just one suggest.

pranithk:
@pkalever @lxbsz I didn't check the code. Here are some doubts about the approach:
1) Is it possible to do the load balancing based on the hash of the gbid instead of maintaining state on the volume? Hash present in glusterfs dht gives near balanced results.
2) How does regen config work when we add these changes? Should it also handle this part and generate the prio related configuration?

pranithk:
> What we have implemented is a counter based logic, block1 on NODE1, block2 on NODE2, block3 on NODE3, block4 on NODE1 and goes on, which fairly gives the true balancing. Do you foresee any issues with this simple logic ? and how does the Hash based algorithm play better ?

There is nothing wrong with the round-robin implementation you did, just that it necessitates that we maintain state. If we can do something equivalent without maintaining state, it is easier to maintain. We don't need to worry too much about keeping correct metadata after replace-block, replace-node etc etc. Every time we add a command that does something that affects active path, we need to worry about updating metadata in this implementation. With stateless solution, we don't need to worry about keeping correct metadata. That's the only reason.

By the way, how do we handle for replace-block and replace-node for the active paths?

pkalever:
> By the way, how do we handle for replace-block and replace-node for the active paths?
I did not handle the replace block yet, but it is very easy, we can handle this from meta-info, i.e. if we have to replace the NODE which matches "PRIOPATH: NODEIP", we need to make it PRIOPATH, that is all. 

Thanks!

pranithk:
@pkalever I didn't understand if you agree about the hash based solution or not. Let us not introduce extra metadata when we can implement the solution without it.

pkalever:
@pranithk @lxbsz The new patch uses getxattr/setxattr way to store prio count, please take a look.

Ex:
$ for i in {1..4}; do \
  gluster-block create sample/block${i} ha 3 \
   192.168.124.149,192.168.124.162,192.168.124.184 1GiB; \
done

$ getfattr -m. -d -e text /mnt/block-meta/prio.info
[...]
user.block.192.168.124.149="2\0..."
user.block.192.168.124.162="1\0..."
user.block.192.168.124.184="1\0..."

While we delete a block, counter for respective prio portal is decremented

$ gluster-block delete sample/block4

$ getfattr -m. -d -e text /mnt/block-meta/prio.info
[...]
user.block.192.168.124.149="1\0..."
user.block.192.168.124.162="1\0..."
user.block.192.168.124.184="1\0..."


While the replace node and capability support is yet to be added.

Thanks!

pkalever:
@lxbsz @pranithk Updated the patch with, replace node and capability support. Please take a look.

NOTE: I haven't got a chance to test it myself, will update with my test results once done.

Thanks!

pkalever:
@lxbsz @pranithk my basic test validation is positive. Please go ahead with the review and further testing.

lxbsz:
@pkalever One thing let's have a record here as we discussed yesterday.
In the future let's create one separate tpg for per path/node and this will benefit from the STPG ALUA feature. STPG will allow the client/initiator to control and set the active/optimized path as it likes.

Currently two tpgs should be enough.

I will test review and test it today.

pkalever:
> @pkalever One thing let's have a record here as we discussed yesterday.
> In the future let's create one separate tpg for per path/node and this will benefit from the STPG ALUA feature. STPG will allow the client/initiator to control and set the active/optimized path as it likes.
> 

Sure, lets take this up in the next release.

> Currently two tpgs should be enough.
> 
> I will test review and test it today.

all right, thanks @lxbsz

lxbsz:
@pkalever 

BTW, to support the failover in tcmu-runner, which patch you are using ? https://github.com/open-iscsi/tcmu-runner/pull/416 ? or the changes Mike has mentioned in one of the comments ? 

Thanks,

lxbsz:
@pkalever 

Test it and it works well for me.

[root@gblock1 block-meta]# ls
block0  block1  block10  block11  block2  block3  block4  block5  block6  meta.lock  prio.info
[root@gblock1 block-meta]# pwd
/mnt/block-meta
[root@gblock1 block-meta]# attr -l prio.info 
Attribute "block.192.168.195.135" has a 1024 byte value for prio.info
Attribute "block.192.168.195.134" has a 1024 byte value for prio.info
[root@gblock1 block-meta]# attr -g "block.192.168.195.135" prio.info 
Attribute "block.192.168.195.135" had a 1024 byte value for prio.info:
4
[root@gblock1 block-meta]# attr -g "block.192.168.195.134" prio.info 
Attribute "block.192.168.195.134" had a 1024 byte value for prio.info:
3
[root@gblock1 block-meta]#

lxbsz:
@pkalever 
It looks good to me.

rpc/block_svc_routines.c
```diff
@@ -1916,22 +1917,26 @@ glusterBlockReplaceNodeRemoteAsync(struct glfs *glfs, blockReplaceCli *blk,
...
+        /* Construct block_hosts */
         tmp = cobj->block_hosts;
         if (GB_ASPRINTF(&cobj->block_hosts, "%s,%s", tmp, info->list[i]->addr) == -1) {
           GB_FREE (tmp);
           goto out;
         }
         GB_FREE(tmp);
+
+        /* Get prio idx */
+        if (info->prio_path && !strcmp(info->prio_path, info->list[i]->addr)) {
```
pkalever: this should be if (info->prio_path[0] && ...

Will fix only this as part of next update.

pkalever:
@lxbsz @pranithk updated the PR, with self comments above.

lxbsz:
@pkalever 
The depended tcmu-runner patch has been upstreamed and merged, please see:
https://github.com/open-iscsi/tcmu-runner/pull/424

rpc/block_svc_routines.c
```diff
@@ -1621,6 +1621,7 @@ glusterBlockBuildMinCaps(void *data, operations opt)
     if (cblk->rb_size) {
       minCaps[GB_CREATE_RING_BUFFER_CAP] = true;
     }
+    minCaps[GB_CREATE_LOAD_BALANCE_CAP] = true;
```
pranithk: Will this be backward compatible if we set this to true by default?

rpc/block_svc_routines.c
```diff
@@ -1916,22 +1917,26 @@ glusterBlockReplaceNodeRemoteAsync(struct glfs *glfs, blockReplaceCli *blk,
...
         tmp = cobj->block_hosts;
         if (GB_ASPRINTF(&cobj->block_hosts, "%s,%s", tmp, info->list[i]->addr) == -1) {
           GB_FREE (tmp);
           goto out;
         }
         GB_FREE(tmp);
+
+        /* Get prio idx */
+        if (info->prio_path[0] && !strcmp(info->prio_path, info->list[i]->addr)) {
+          cobj->prio_idx = j;
```
pranithk: It is better to send the prio_path instead of prio index over the wire as it will be independent of the way info is constructed as of today. In future if there is a change in the way meta is constructed which changes the order of hosts then this code can lead to problem?

rpc/rpcl/block.x
```diff
@@ -32,6 +32,7 @@ struct blockCreate2 {
   char      passwd[127];                 /* uuid */
   u_quad_t  size;
   char      block_name[255];
+  u_int     prio_idx;
```
pranithk: It is better to send the ip as mentioned before.

rpc/glfs-operations.c
```diff
@@ -674,3 +677,138 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+  return;
+}
+
+
+void
+blockDecPrioAttr(struct glfs* glfs, char *volume, char *addr)
+{
+  size_t count;
+  char buf[1024] = {'\0', };
+  char attr[256 + 12] = {'\0', };
```
pranithk: xattr key length has a limit of 255 chars. Value has a limit of 64KB. http://man7.org/linux/man-pages/man5/attr.5.html So this needs to change.

rpc/glfs-operations.c
```diff
@@ -674,3 +677,138 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+  return index + 1;
+}
+
+
+void
+blockIncPrioAttr(struct glfs* glfs, char *volume, char *addr)
+{
+  size_t count;
+  char buf[1024] = {'\0', };
+  char attr[256 + 12] = {'\0', };
```
pranithk: xattr key length has a limit of 255 chars. Value has a limit of 64KB. http://man7.org/linux/man-pages/man5/attr.5.html So this needs to change.

rpc/glfs-operations.c
```diff
@@ -674,3 +677,138 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+ out:
+  if (pfd && glfs_close(pfd) != 0) {
+    LOG("gfapi", GB_LOG_ERROR, "glfs_close(%s): on volume %s failed[%s]",
+        GB_PRIO_FILE, volume, strerror(errno));
+  }
+
+  if (ret)
+    return 1;        /* on failure let the prio index be 1 */
+
+  return index + 1;
```
pranithk: In general index starts from '0', is there any reason why we are not deviating from that?

rpc/block_svc_routines.c
```diff
@@ -3451,14 +3460,16 @@ block_create_cli_1_svc_st(blockCreateCli *blk, struct svc_req *rqstp)
     goto exist;
   }
 
+  cobj.prio_idx = blockGetPrioPathIndex(glfs, blk->volume, list);
```
pranithk: Shouldn't the prio calculations done based on whether the other servers in the cluster have this capability or not?

rpc/block_svc_routines.c
```diff
@@ -3451,14 +3460,16 @@ block_create_cli_1_svc_st(blockCreateCli *blk, struct svc_req *rqstp)
...
+
   uuid_generate(uuid);
   uuid_unparse(uuid, gbid);
 
   GB_METAUPDATE_OR_GOTO(lock, glfs, blk->block_name, blk->volume,
                         errCode, errMsg, exist,
                         "VOLUME: %s\nGBID: %s\n"
-                        "HA: %d\nENTRYCREATE: INPROGRESS\n",
-                        blk->volume, gbid, blk->mpath);
+                        "HA: %d\nENTRYCREATE: INPROGRESS\nPRIOPATH: %s\n",
```
pranithk: same as above

lxbsz:
@pkalever

This is the multipath.conf about this, please review.

devices {
        device {
                vendor "LIO-ORG"
                user_friendly_names "yes" # names like mpatha
                path_grouping_policy "failover" # one path per group
                hardware_handler "1 alua"
                path_selector "round-robin 0"
                failback immediate
                path_checker "tur"
                prio "alua"
                no_path_retry 120
        }
}

Thanks,

rpc/block_svc_routines.c
pkalever: If someone is trying to Create from an updated gluster-block node, create will fail if one of the other nodes are not having load-balancing capability. Do we want to allow create to successed with this case ?

rpc/block_svc_routines.c
pkalever: Sure.

rpc/glfs-operations.c
pkalever: Okay I will defend on 255 char max and fail if it exceeds the range.

rpc/glfs-operations.c
pkalever: As we want to pass IP this will go off now.

rpc/block_svc_routines.c
pkalever: That is way we fail if capabilities mismatch, hence the question above.

pkalever:
Right now we are failing on create request if capabilities mismatch. I would like to know your opinion about this.

Thanks @pranithk

rpc/block_svc_routines.c
pranithk: If there are gluster-blockds without load-balancing support in the cluster, we need to downgrade the create operation to be done without load-balancing and succeed the operation.

pkalever:
@pranithk updated PR with the review comments.

Thanks!

rpc/block_svc_routines.c
```diff
@@ -3451,14 +3478,19 @@ block_create_cli_1_svc_st(blockCreateCli *blk, struct svc_req *rqstp)
...
+                        blk->volume, gbid, blk->mpath,
+                        cobj.prio_path[0] ? cobj.prio_path : "");
```
pranithk: PRIOPATH: line itself shouldn't come when prio_path doesn't exist for backward compatibility right?

rpc/glfs-operations.c
```diff
@@ -674,3 +678,142 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
   return ret;
 }
+
+
+void
+blockGetPrioPath(struct glfs* glfs, char *volume, blockServerDefPtr list,
+                 char *prio_path, size_t prio_len)
+{
+  struct glfs_fd *pfd = NULL;
+  char attr[255];
```
pranithk: This should be 256. The resultant key length will be 255.

rpc/glfs-operations.c
```diff
@@ -674,3 +678,142 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+                   S_IRUSR | S_IWUSR);
+  if (!pfd) {
+    LOG("gfapi", GB_LOG_ERROR, "glfs_creat(%s) on volume %s failed[%s]",
+        GB_PRIO_FILE, volume, strerror(errno));
+    return;
+  }
+
+  for (i = 0; i < list->nhosts; i++) {
+    memset(attr, '\0', 255);
+    snprintf(attr, 255, "%s.%s", GB_LB_ATTR_PREFIX, list->hosts[i]);
```
pranithk: Better to change above 2 instances to sizeof(attr)

rpc/glfs-operations.c
```diff
@@ -674,3 +678,142 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+  return;
+}
+
+
+void
+blockIncPrioAttr(struct glfs* glfs, char *volume, char *addr)
+{
+  size_t count;
+  char buf[1024] = {'\0', };
+  char attr[255] = {'\0', };
```
pranithk: This should be 256

rpc/glfs-operations.c
```diff
@@ -674,3 +678,142 @@ blockGetMetaInfo(struct glfs* glfs, char* metafile, MetaInfo *info,
...
+
+
+  snprintf(attr, 255, "%s.%s", GB_LB_ATTR_PREFIX, addr);
```
pranithk: use sizeof(attr)

rpc/block_svc_routines.c
pkalever: May be, you are right.

pkalever:
@pranithk Updated with the latest review comments now. Please review.

PR #89, opened 2018-06-08
Crash fix

pkalever:
* fill errMsg and delete metafile on EntryCreate fail
* fix other compiler warnings

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -571,6 +571,7 @@ glusterBlockCallRPC_1(char *host, void *cobj,
   case LIST_SRV:
   case INFO_SRV:
   case REPLACE_GET_PORTAL_TPG_SRV:
+  case GENCONFIG_SRV:
       goto out;
   case REPLACE_SRV:
       *rpc_sent = TRUE;
@@ -1726,6 +1727,7 @@ glusterBlockBuildMinCaps(void *data, operations opt)
   case LIST_SRV:
   case INFO_SRV:
   case VERSION_SRV:
+  case GENCONFIG_SRV:
     break;
   }
 
@@ -2591,7 +2593,7 @@ getTpgObj(char *block, MetaInfo *info, blockGenConfigCli *blk, char *portal, int
   uuid_generate(uuid);
   uuid_unparse(uuid, alias);
   snprintf(alias_10, 11, "%.10s", alias+24);
-  json_object_object_add(tpg_lun_obj, "alias", GB_JSON_OBJ_TO_STR(alias_10));
+  json_object_object_add(tpg_lun_obj, "alias", GB_JSON_OBJ_TO_STR(alias_10[0]?alias_10:NULL));
   snprintf(lun_so, 256, "/backstores/user/%s", block);
   if (info->prio_path[0]) {
     if (!strcmp(info->prio_path, portal)) {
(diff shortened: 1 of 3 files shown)


pkalever:
@pranithk @lxbsz please review.

Thanks!

rpc/block_svc_routines.c
```diff
@@ -2591,7 +2593,7 @@ getTpgObj(char *block, MetaInfo *info, blockGenConfigCli *blk, char *portal, int
   uuid_generate(uuid);
   uuid_unparse(uuid, alias);
   snprintf(alias_10, 11, "%.10s", alias+24);
-  json_object_object_add(tpg_lun_obj, "alias", GB_JSON_OBJ_TO_STR(alias_10));
+  json_object_object_add(tpg_lun_obj, "alias", GB_JSON_OBJ_TO_STR(alias_10[0]?alias_10:NULL));
```
lxbsz: Are this and the following changes really needed here ?

pranithk:
@pkalever Could you add details about what was the problem explaining the reason for crash and a little bit more about the fix to the commit message?

rpc/block_svc_routines.c
pkalever: These are to avoid gcc and rpmbuild warnings. Are you saying we should make this as a separate patch/PR ?

pkalever:
@pranithk Sure.

pkalever:
@pranithk @lxbsz updated the PR. Thanks!

rpc/block_svc_routines.c
lxbsz: I test the old code, couldn't see any warning info of this. I thought the alias_10[0] couldn't be '\0' because of the last line of the code would fill something in it.

pkalever: Nice. Could you try:

$ make clean
$ make CFLAGS="-Wall"

Thanks!

lxbsz: Cool.

pranithk:
@pkalever glusterBlockCreateEntry() seems to already have catch-all errMsg creation:
```

 if (!errMsg) {                                                               
 GB_ASPRINTF (errMsg, "Not able to create storage for %s/%s [%s]",          
                     blk->volume, blk->block_name, strerror(*errCode));            
 }  
```

Do we know why that didn't take effect?

lxbsz:
> @pkalever glusterBlockCreateEntry() seems to already have catch-all errMsg creation:
> 
> 
>  if (!errMsg) {                                                               
>  GB_ASPRINTF (errMsg, "Not able to create storage for %s/%s [%s]",          
>                      blk->volume, blk->block_name, strerror(*errCode));            
>  }  
> Do we know why that didn't take effect?

I think here we should check:
```
if (errMsg && !(*errMsg))
...
```

pkalever:
@pranithk @lxbsz Good catch, I did not looked at glusterBlockCreateEntry(). I have caught this on a gdb session with the core file, just fixed it this way.

Thanks for the insight. Will fix it this way.

pranithk:
Ah! @lxbsz good catch

pkalever:
@pranithk @lxbsz updated it as per the suggestions.

Thanks!

lxbsz:
@pkalever 
I was running the test  script from last night till now, haven't reproduced the crash.
And the the code looks good to me.

Thanks,

rpc/block_svc_routines.c
```diff
@@ -3900,8 +3899,12 @@ block_create_cli_1_svc_st(blockCreateCli *blk, struct svc_req *rqstp)
   }
 
   if (glusterBlockCreateEntry(glfs, blk, gbid, &errCode, &errMsg)) {
-    LOG("mgmt", GB_LOG_ERROR, "%s volume: %s host: %s",
-        FAILED_CREATING_FILE, blk->volume, blk->block_hosts);
+    LOG("mgmt", GB_LOG_ERROR, "%s volume: %s block: %s file: %s host: %s",
+        FAILED_CREATING_FILE, blk->volume, blk->block_name, gbid, blk->block_hosts);
+    if (glusterBlockDeleteMetaFile(glfs, blk->volume, blk->block_name)) {
```
pranithk: Even metafile deletion is happening in the  glusterBlockCreateEntry() function? Could you also move the LOG inside that function? May be in another patch it is better to remove glfs_chdir() usage from glfs-operations.c, that way we don't have to worry about pwd ever.

PR #92, opened 2018-07-02
gluster-block: set the cmd_time_out to 130

lxbsz:
0 means the fired or pending SCSI commands will wait forever, and
when the IOs going on and tcmu-runner process is crashed, so the
iscsi_ttx process will stucked forever.

Fixes: bz#1596684

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -4243,7 +4243,7 @@ block_create_common(blockCreate *blk, char *rbsize, char *prio_path)
   }
 
   if (GB_ASPRINTF(&backstore_attr,
-                  "%s/%s set attribute cmd_time_out=0",
+                  "%s/%s set attribute cmd_time_out=130",
                   GB_TGCLI_GLFS_PATH, blk->block_name) == -1) {
     goto out;
   }


pkalever:
setting cmd_time_out to 0 means that the iscsi_trx will wait infinite until the cms fired to tcmu-runner return back. Just in case if tcmu-runner is killed/terminated, the iscsi_trx will hung in D state infinitely, which later result in KERNEL WARNING messages, like:

(2018-07-02 2:44:41 PM) xiubli_2: Jul 02 04:38:34 rhel1 kernel: Call Trace:
(2018-07-02 2:44:43 PM) xiubli_2: Jul 02 04:38:34 rhel1 kernel: iscsi_ttx       D ffff9a69b6a64100     0  3413      2 0x00000084
[...]


The defaults,
GLUSTER ping timeout = 42
KERNEL warnings timeout  = 120

so cmd_time_out should be 
42s < cmd_time_out < 120s

Setting cmd_time_out = 60sec should be good.

Thanks!

lxbsz:
@pkalever  

Update the commit comment with your last comment.
Thanks very much.

lxbsz:
@pkalever @pranithk 
For this PR, please wait and don't merge it, there maybe new situation about bz#1596684, we may switch to add reset ring support instead of this.

lxbsz:
@pkalever @pranithk 
This still needed to fix the bz.

The defaults,
GLUSTER ping timeout = 42
/etc/iscsi/iscsi.conf:node.session.timeo.replacement_timeout = 120

so cmd_time_out should be
cmd_time_out > GLUSTER ping timeout && cmd_time_out > replacement_timeout

The hack is to make cmd_time_out expire after
fast_io_fail_tmo/replacement_timeout. This way the path will be
marked bad by multipathd and the IO failed due to cmd_time_out
firing will be retried on another path instead of the one where
runner is down. Path tester IO will still get sent to the down
path but that will not disrupt IO on good paths.

Though the D state could recovery from LIO's reset ring option,
this is still needed.

We must make sure that the customer's setting of replacement_timeout
value < 180 and tell customer that when this bz is hit, there will
be 3 mintues' stuck of gluster-block/tcmu-runner node.

pkalever:
@lxbsz agree and and this 3 min wait will be solved by reset ring patch.

In absence of reset ring fix, one has to wait for 3 min's before restarting gluster-block-target, in-order to recover from the D state hit earlier.

pkalever:
As discussed, since the default replacement timeout is 120, and I do not expect our users to change this to > 120 ATM. bringing our cmd_time_out to 130 (>120) from 180 will save us some time.

@lxbsz I will leave the discussion to you.

lxbsz:
> As discussed, since the default replacement timeout is 120, and I do not expect our users to change this to > 120 ATM. bringing our cmd_time_out to 130 (>120) from 180 will save us some time.
> 
> @lxbsz I will leave the discussion to you.

Yeah, agree and 130s will be okay, I didn't foresee any use case which will change the default replacement timeout.

pkalever:
Can you please add the reviewed-by tags in the 'Reviewed-by:' tags in the patch ?
Once you are done that, I will hit the merge button.

lxbsz:
@pkalever 
Done.

pkalever:
Looks good to me, Thanks @lxbsz

pkalever:
@lxbsz I think I have jumped the gun. Could you add cmd_time_out at part of genconfig as well ?

PR #94, opened 2018-07-06
delete: fix glusterBlockDeleteRemoteAsync ret value

pkalever:
Before:
[root@localhost gluster-block]# gluster-block delete sample/block
FAILED ON:   192.168.124.217
SUCCESSFUL ON:   192.168.124.6 192.168.124.163
RESULT: SUCCESS

After:
[root@localhost gluster-block]# gluster-block delete sample/block
FAILED ON:   192.168.124.217
SUCCESSFUL ON:   192.168.124.163 192.168.124.6
RESULT: FAIL

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -1265,8 +1265,8 @@ glusterBlockDeleteRemoteAsync(char *blockname,
   if (ret) {
     goto out;
   }
-
   ret = -1;
+
   if (d_attempt) {
     a_tmp = local->d_attempt;
     if (GB_ASPRINTF(&local->d_attempt, "%s %s",
@@ -1305,6 +1305,7 @@ glusterBlockDeleteRemoteAsync(char *blockname,
   if (ret) {
     goto out;
   }
+  ret = -1;
 
   for (i = 0; i < info_new->nhosts; i++) {
     switch (blockMetaStatusEnumParse(info_new->list[i]->status)) {
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -41,6 +41,7 @@
                                 "'%s' ls | grep -e tpg -e '%s' | grep -B1 '%s' | grep -o 'tpg\\w'"
 # define   GB_CHECK_PORTAL      "targetcli /iscsi/" GB_TGCLI_IQN_PREFIX \
                                 "'%s' ls | grep '%s' > " DEVNULLPATH
+# define   GB_SAVECONFIG_CHECK  "grep -m 1 '\"name\": \"%s\",' " GB_SAVECONFIG " > " DEVNULLPATH
 
 # define   GB_ALUA_AO_TPG_NAME          "glfs_tg_pt_gp_ao"
 # define   GB_ALUA_ANO_TPG_NAME         "glfs_tg_pt_gp_ano"
@@ -54,6 +55,7 @@
 # define   GB_NODE_NOT_EXIST    223
 # define   GB_NODE_IN_USE       224
 # define   GB_BLOCK_NOT_LOADED  225
+# define   GB_BLOCK_NOT_FOUND   226
 
 pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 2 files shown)


pkalever:
@pranithk @lxbsz please review

pkalever:
@pranithk as you are aware,


We have covered case:
-----------------------

tree 04238f51403458e3fcf4906ef56e7fc425e44b95                                    
parent 0b0b7db14159e009d60435c38ded42a5606472ad                                  
author Prasanna Kumar Kalever <prasanna.kalever@redhat.com> Tue Jun 5 13:39:31 2018 +0530
committer Pranith Kumar Karampuri <pranith.k@gmail.com> Tue Jun 5 19:35:58 2018 +0800
                                                                                 
delete: defend on block not loaded and CLEANUPFAIL status                        
                                                                                 
Consider the following case:                                                     
                                                                                 
* When gluster-block delete was executed for the first time, on this node,       
  deletion was successful. But before it can send the response,                  
  gluster-blockd died (addr meta status would be CLEANUPFAIL)                    
                                                                                 
* When the user attempts gluster-block deletion, it should succeed.              
                                                                                                                                                                                                                   
With above change, if the block is not loaded on 1 node out of 3, the first delete will fail saying block not loaded in the logs and the second delete will succeed. Are we okay with this ?

pranithk:
@pkalever Please add the problem statement and what you are trying to fix.

pranithk:
> With above change, if the block is not loaded on 1 node out of 3, the first delete will fail saying block not loaded in the logs and the second delete will succeed. Are we okay with this ?

@pkalever It is not a good user experience to fail and then succeed.

pkalever:
@pranithk @lxbsz please review.

Issue https://github.com/gluster/gluster-block/issues/96 is to track the Hints fix, for now will just log.

rpc/block_svc_routines.c
```diff
@@ -258,6 +260,67 @@ blockStr2arrayAddToJsonObj(json_object *json_obj, char *string, char *label)
...
+    LOG("mgmt", GB_LOG_ERROR, "%s", reply->out);
+    goto out;
+  } else if (ret == 1) {
+    is_loaded = false;
+    GB_ASPRINTF(&reply->out, "Block '%s' may be not loaded\n", block_name);
+    LOG("mgmt", GB_LOG_ERROR, "%s", reply->out);
+  }
+  GB_FREE(exec);
+
+  if (GB_ASPRINTF(&exec, GB_SAVECONFIG_CHECK, block_name) == -1) {
```
pranithk: Please set return value indicating error

rpc/block_svc_routines.c
```diff
@@ -258,6 +260,67 @@ blockStr2arrayAddToJsonObj(json_object *json_obj, char *string, char *label)
...
+  if (GB_ASPRINTF(&exec, GB_TGCLI_CHECK, block_name, gbid) == -1) {
+    goto out;
+  }
+
+  ret = gbRunner(exec);
+  if (ret == -1) {
+    GB_ASPRINTF(&reply->out, "command exit abnormally for %s\n", block_name);
+    LOG("mgmt", GB_LOG_ERROR, "%s", reply->out);
+    goto out;
+  } else if (ret == 1) {
```
pranithk: exit value 0 means success and non-zero is failure. So are we sure that ret value is going to be 1 for this failure? Same comment for the next set of if/else as well

rpc/block_svc_routines.c
```diff
@@ -258,6 +260,67 @@ blockStr2arrayAddToJsonObj(json_object *json_obj, char *string, char *label)
...
+    goto out;
+  }
+
+  ret = gbRunner(exec);
+  if (ret == -1) {
+    GB_FREE(reply->out);
+    GB_ASPRINTF(&reply->out, "command exit abnormally for %s\n", block_name);
+    LOG("mgmt", GB_LOG_ERROR, "%s", reply->out);
+    goto out;
+  } else if (ret == 1 || ret == 2) {  /* search not found or savefile not exist */
```
pranithk: same comment as above

rpc/block_svc_routines.c
pkalever: Good catch.

rpc/block_svc_routines.c
pkalever: Yes, it will be 1 (not something newly introduced)

rpc/block_svc_routines.c
pkalever: Yes, 2 is extra here for file not found case (grep on savefile case)

pkalever:
@pranithk update the ret = -1 on GB_ASPRINTF() fail

Thanks!

rpc/block_svc_routines.c
pranithk: Isn't it better to flip the condition? If ret == 0 consider success otherwise consider failure? Isn't that more robust?

pkalever:
Added:
 /* if block is loaded, skip the rest */
+  if (!ret) {
+    goto out;
+  }

pkalever:
Added:

```
if (list->nhosts != info->mpath) {                                      
  errCode = ENOENT;                                                     
                                                                        
  GB_ASPRINTF(&errMsg, "Some of the nodes are missing configuration, " \
              "please look at logs for recent operations on the %d.",   
              blk->block_name);                                         
  LOG("mgmt", GB_LOG_ERROR, "%s", errMsg);                              
                                                                        
  goto out;                                                             
}                                                                       

```

Case: when delete is issued, in case if there are 2 nodes where delete succeeded and on 1 node, block not loaded. It was returning success.

rpc/block_svc_routines.c
pkalever: will generalize this part later, may be in a different revision as this is well tested lets live with it for now :-)

pkalever:
I have tested this patch PR with various cases like:

1. block not loaded, but exist in the saveconfig.json
result: delete and modify both fail
2. block loaded, but not exist in the saveconfig.json
result: delete and modify both success
3. block loaded, but saveconfig.json file missing
result: delete and modify both success
4. block not loaded and not exist in the saveconfig.json
result: delete success and modify fail
5. block deleted on one node, then the subsequent modify will error out with 
"errMsg":"Some of the nodes are missing configuration, please look at logs for recent operations on the 'block1'."

Thanks!

pkalever:
@pranithk tags refreshed.

PR #98, opened 2018-07-11
Script to wait for bricks to come online

pranithk:
Problem:
When a node is rebooted, by the time tcmu-runner starts there is no
guarantee that the bricks of the volume on which the block-volumes
are created will be online. This leads to a case where doing open
on the block files leads to open failing with ENOTCONN leading
to containers with this as pvc are in failed state.

Fix:
A script is added which will be added as a unit file dependency for
gluster-blockd which makes sure that the bricks are online for the
block-hosting-volume preventing this situation.

### DIFF
--- a/extras/wait-for-bricks.sh
+++ b/extras/wait-for-bricks.sh
@@ -1,0 +1,162 @@
+#!/bin/bash
+#Copyright (c) 2018 Red Hat, Inc. <http://www.redhat.com>
+#This file is part of gluster-block.
+#
+#This file is licensed to you under your choice of the GNU Lesser
+#General Public License, version 3 or any later version (LGPLv3 or
+#later), or the GNU General Public License, version 2 (GPLv2), in all
+#cases as published by the Free Software Foundation.
+
+# This file waits for gluster block hosting volume bricks to be up. If all the
+# bricks are up, it exits with success status, otherwise waits for the $timeout
+# seconds before erroring out with the status about
+# online-bricks/expected-online-bricks
+
+LOGDIR="${GB_LOGDIR:-/var/log/gluster-block}"
+mkdir -p "${LOGDIR}"
+
+function printLog()
+{
+    local msg=${1}
+
+    echo "[$(date -u +'%Y-%m-%d %I:%M:%S')] ${msg}" >> "${LOGDIR}/gluster-block-bricks-start.log"
+}
+
+function volinfo_field()
+{
+    local vol=$1;
+    local field=$2;
+
+    gluster volume info "$vol" --xml | grep -oPm1 "(?<=<$field>)[^<]+"
+}
+
+function is_enabled()
+{
+        case "$1" in
+        0|off|no|false|disable)
+                echo "N"
+                ;;
+        1|on|yes|true|enable)
+                echo "Y"
+                ;;
+        *)
+                echo "N/A"
+                ;;
+        esac
+}
+
+function volinfo_option()
+{
+    local vol=$1;
+    local key=$2;
+
+    gluster volume get "$vol" "$key"| awk '{print $NF}' | tail -1
+}
+
+function check_brick_status() {
+        local volname=$1
+        local daemon=$2
+
+        cmd="gluster --xml volume status $volname"
+        if [[ -z "$daemon" ]]
+        then
+                 $cmd | grep -c '<status>1'
+        else
+                 $cmd | sed -n -e "/${daemon}/,/<status>/ p" | grep -c '<status>1'
+        fi
+}
+
+function volume_online_brick_count()
+{
+        local volname=$1
+        local v1=0
+        local v2=0
+        local v3=0
+        local v4=0
+        local v5=0
+        local tot=0
+
+        #First count total Number of bricks and then subtract daemon status
+        v1=$(check_brick_status "$volname")
+        v2=$(check_brick_status "$volname" "Self-heal")
+        v3=$(check_brick_status "$volname" "Quota")
+        v4=$(check_brick_status "$volname" "Snapshot")
+        v5=$(check_brick_status "$volname" "Tier")
+        tot=$((v1-v2-v3-v4-v5))
+        echo $tot
+}
+
+function volumes_waiting_for_bricks_up()
+{
+        local wait_info
+        local vol_down_info
+        while read -r volname
+        do
+                if [[ "Started" == "$(volinfo_field "$volname" 'statusStr')" ]]
+                then
+                        if [[ "Y" == "$(is_enabled "$(volinfo_option "$volname" features.shard)")" ]]
+                        then
+                                brick_count="$(volume_online_brick_count "$volname")"
+                                total_brick_count=$(volinfo_field "$volname" 'brickCount')
+                                if [[ $brick_count -ne $total_brick_count ]]
+                                then
+                                        vol_down_info="$volname ($((total_brick_count - brick_count))/$total_brick_count)"
+                                        if [[ -z "$wait_info" ]]
+                                        then
+                                                wait_info="$vol_down_info"
+                                        else
+                                                wait_info="$wait_info, $vol_down_info"
+                                        fi
+                                fi
+                        fi
+                fi
+        done < <(gluster volume list)
+        if [[ ! -z "$wait_info" ]]; then echo "$wait_info"; fi
+}
+
+function check_glusterd_running()
+{
+        if ! pidof glusterd > /dev/null 2>&1
+        then
+                printLog "ERROR: Glusterd is not running";
+                exit 1;
+        fi
+}
+
+function wait_for_bricks_up()
+{
+        local endtime=$1
+        local now
+
+        while [[ ! -z "$(volumes_waiting_for_bricks_up)" ]]
+        do
+                now=$(date +%s%N)
+                if [[ $now -ge $endtime ]]
+                then
+                        printLog "WARNING: Timeout Expired, bricks of volumes:\"$(volumes_waiting_for_bricks_up)\" are yet to come online"
+                        exit 1
+                fi
+                sleep 1;
+        done
+}
+
+if [[ "$#" -ne 1 ]]
+then
+        echo "Usage $0 <timeout-in-seconds>"
+        exit 1
+fi
+
+case $1 in
+    ''|*[!0-9]*) echo "Usage $0 <timeout-in-seconds>"; exit 1 ;;
+    *) ;;
+esac
+
+check_glusterd_running;
+timeout=$1
+start=$(date +%s%N)
+#Convert timeout to Nano Seconds as 'start' is in Nano seconds
+endtime=$(( timeout*1000000000 + start ))
+wait_for_bricks_up $endtime
+now=$(date +%s%N)
+printLog "INFO: All bricks for Block Hosting Volumes came online in $(((now-start)/1000000000)) seconds"
+exit 0
--- a/tests/glusterfs/wait-for-bricks-test.sh
+++ b/tests/glusterfs/wait-for-bricks-test.sh
@@ -1,0 +1,99 @@
+#!/bin/bash -x
+
+#This test-script tests that wait-for-bricks.sh works as expected in different cases
+exit_code=0
+source_path="../../extras"
+
+function test_success {
+        if ! "$@" ;
+        then
+                exit_code=1
+        fi
+}
+
+function test_failure {
+        if "$@" ;
+        then
+                exit_code=1
+        fi
+}
+
+#Should fail when glusterd is not running
+test_failure $source_path/wait-for-bricks.sh 10
+
+#Should succeed when gluster --mode=scriptd is running but no volumes exist
+glusterd
+test_success $source_path/wait-for-bricks.sh 10
+
+#Set brick-multiplexing off, so that we can kill bricks.
+gluster volume set all cluster.brick-multiplex off
+
+#Block hosting volumes are identified with features.shard being present for now
+#Should succeed when volumes that are not of interest are present
+gluster --mode=script volume create vol localhost.localdomain:/brick1 force
+test_success $source_path/wait-for-bricks.sh 10
+
+gluster --mode=script volume start vol
+test_success $source_path/wait-for-bricks.sh 10
+
+gluster --mode=script volume create vol2 replica 3 localhost.localdomain:/brick{2..4} force
+test_success $source_path/wait-for-bricks.sh 10
+
+gluster --mode=script volume start vol2
+test_success $source_path/wait-for-bricks.sh 10
+
+#Checking for the case when only the volume with replicate is available
+gluster --mode=script volume stop vol
+test_success $source_path/wait-for-bricks.sh 10
+gluster --mode=script volume start vol
+
+#Checking for the case where bhv is present but stopped
+gluster --mode=script volume create bhv1 replica 3 localhost.localdomain:/brick{5..7} force
+gluster --mode=script volume set bhv1 features.shard on
+test_success $source_path/wait-for-bricks.sh 10
+
+#Start bhv1 so that wait-for-bricks considers this volume
+gluster --mode=script volume start bhv1
+test_success $source_path/wait-for-bricks.sh 10
+
+#When a brick in bhv1 is down, script should fail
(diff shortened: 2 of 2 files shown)


extras/wait-for-bricks-test.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash -x
```
pkalever: Should we disable trace output (-x) ? or is this intentional ?

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/bash
+#Copyright (c) 2018 Red Hat, Inc. <http://www.redhat.com>
+#This file is part of GlusterFS.
```
pkalever: s/GlusterFS/gluster-block/ ?

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+                fi
+        done | tail -1
+}
+
+function check_brick_status() {
+        local volname=$1
+        local daemon=$2
+
+        cmd="gluster --xml volume status $volname"
+        if [[ -z $daemon ]]
```
pkalever: If we want to use [[    ]]  style lets keep it consistent through out the file. But from my past experience we better use [[   ]] instead of [  ]

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+                        exit 1
+                fi
+                sleep 1;
+        done
+}
+
+check_glusterd_running;
+timeout=$1
+start=$(date +%s%N)
+endtime=$(( ${timeout}*1000000000 +start ))
```
pkalever: May be you want $(( ${timeout}*1000000000 ) + start ) ?

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+wait_for_bricks_up $endtime
+now=$(date +%s%N)
+echo "All bricks for Block Hosting Volumes came online in $(((now-start)/1000000000)) seconds"
+exit 0;
```
pkalever: Why we had to return success at the end ? It make-sense to fail in case bricks won't come by timeout IMO.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+        local volname=$1
+        local v1=0
+        local v2=0
+        local v3=0
+        local v4=0
+        local v5=0
+        local tot=0
+
+        #First count total Number of bricks and then subtract daemon status
+        v1=`check_brick_status $volname`
```
pkalever: To keep code consistent, either we adopt `cmd` or $(cmd). my choice would be the second one.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+#later), or the GNU General Public License, version 2 (GPLv2), in all
+#cases as published by the Free Software Foundation.
+
+
+function volinfo_field()
+{
+    local vol=$1;
+    local field=$2;
+
+    gluster volume info $vol --xml| grep -oPm1 "(?<=<$field>)[^<]+"
```
pkalever: May be Missing space before |

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,159 @@
...
+                return 0
+        fi
+        gluster volume list | while read volname;
+        do
+                if [ "Started" == $(volinfo_field $volname 'statusStr') ];
+                then
+                        if [ "Y" == $(is_enabled $(volinfo_option $volname features.shard)) ];
+                        then
+                                brick_count=$(volinfo_field $volname 'brickCount')
+                                number_of_bricks=$((number_of_bricks + brick_count))
```
pkalever: May be I'm missing something, but why we had to use (( instead of (

pkalever:
@pranithk @amarts @lxbsz @nixpanic 
I'm not sure what should be a right place (path) to keep wait-for-bricks.sh ?
Should we create a new path '/var/lib/gluster-block/' ?

If so, it will be a good opportunity for '/etc/sysconfig/gluster-block-caps.info' to move under '/var/lib/gluster-block/' as well.

Thanks!

extras/wait-for-bricks-test.sh
pkalever: Also should we move the test script under tests/ dir ?

extras/wait-for-bricks.sh
lxbsz: IMO, return success at the end here will be okay. And in the case bricks won't come by timeout just print out the error or warning logs should be okay.

Because there maybe some commands executing errors in the script, and the success at the end means there has no errors for the script itself.

pkalever: Oh! Looks like, we are 'exit 1' at line 146, I think this should be fine then. Sorry my fault.

extras/wait-for-bricks-test.sh
pranithk: It is better for the test to give as much information as possible, so it seemed okay. This test is not testing functionality of gluster-block and testing something in extras. So I kept it that way.

extras/wait-for-bricks.sh
pranithk: Done

extras/wait-for-bricks.sh
nixpanic: Please run this script through `shellcheck`, there are a number of things that the tool suggests to improve.

extras/wait-for-bricks.sh
nixpanic: `$(...)` is generally preferred as it improves readability.

extras/wait-for-bricks.sh
pkalever: Yay!! I learned something new :-)

pranithk:
@pkalever Here are the logs it printed when I tested it:
21:23:47 :( ⚡ tail -f /var/log/gluster-block/gluster-block-bricks-start.log 
[2018-07-16 03:53:40] ERROR: Glusterd is not running
[2018-07-16 03:53:42] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:43] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:44] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:45] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:47] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:49] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-16 03:53:50] INFO: All bricks for Block Hosting Volumes came online in 1 seconds
[2018-07-16 03:53:53] INFO: All bricks for Block Hosting Volumes came online in 1 seconds
[2018-07-16 03:54:07] WARNING: Timeout Expired, online/expected-online brick-counts: 2/3
[2018-07-16 03:54:10] INFO: All bricks for Block Hosting Volumes came online in 1 seconds
[2018-07-16 03:54:12] INFO: All bricks for Block Hosting Volumes came online in 1 seconds
[2018-07-16 03:54:15] INFO: All bricks for Block Hosting Volumes came online in 2 seconds
[2018-07-16 03:54:30] WARNING: Timeout Expired, online/expected-online brick-counts: 5/6
[2018-07-16 03:54:46] WARNING: Timeout Expired, online/expected-online brick-counts: 5/6
[2018-07-16 03:54:49] INFO: All bricks for Block Hosting Volumes came online in 2 seconds

nixpanic:
On Mon, Jul 16, 2018 at 08:57:54AM -0700, Pranith Kumar Karampuri wrote:
> @pkalever Here are the logs it printed when I tested it:
> 21:23:47 :( ⚡ tail -f /var/log/gluster-block/gluster-block-bricks-start.log 
..
> [2018-07-16 03:54:46] WARNING: Timeout Expired, online/expected-online brick-counts: 5/6
> [2018-07-16 03:54:49] INFO: All bricks for Block Hosting Volumes came online in 2 seconds

All messages regarding a subset of the volumes should mention the
affected volumes to make troubleshooting easier. Specially the WARNING
messages will benefit from that.

pranithk:
@nixpanic done
[2018-07-17 05:40:00] ERROR: Glusterd is not running
[2018-07-17 05:40:01] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:02] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:02] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:03] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:04] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:06] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:07] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:09] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:20] WARNING: Timeout Expired, bricks of volumes:"bhv1 (1/3)" are yet to come online
[2018-07-17 05:40:22] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:23] INFO: All bricks for Block Hosting Volumes came online in 0 seconds
[2018-07-17 05:40:27] INFO: All bricks for Block Hosting Volumes came online in 1 seconds
[2018-07-17 05:40:39] WARNING: Timeout Expired, bricks of volumes:"bhv2 (1/3)" are yet to come online
[2018-07-17 05:40:51] WARNING: Timeout Expired, bricks of volumes:"bhv1 (1/3), bhv2 (1/3)" are yet to come online
[2018-07-17 05:41:05] WARNING: Timeout Expired, bricks of volumes:"bhv1 (1/3)" are yet to come online
[2018-07-17 05:41:07] INFO: All bricks for Block Hosting Volumes came online in 1 seconds

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+{
+    local vol=$1;
+    local field=$2;
+
+    gluster volume info "$vol" --xml | grep -oPm1 "(?<=<$field>)[^<]+"
+}
+
+function is_enabled()
+{
+        case $1 in
```
nixpanic: Variables used in expressions should be placed between quotes (like `"$1"`) to prevent problems. In case `$1` is an empty string, the script will error out.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+
+    gluster volume get "$vol" "$key"| awk '{print $NF}' | tail -1
+}
+
+function check_brick_status() {
+        local volname=$1
+        local daemon=$2
+
+        cmd="gluster --xml volume status $volname"
+        if [[ -z $daemon ]]
```
nixpanic: `$daemon` should be in quotes `"$daemon"`, or even better `"${daemon}"`.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+        v5=$(check_brick_status "$volname" "Tier")
+        tot=$((v1-v2-v3-v4-v5))
+        echo $tot
+}
+
+function volumes_waiting_for_bricks_up()
+{
+        local wait_info
+        local vol_down_info
+        while read -r volname;
```
nixpanic: the `;` is not needed at the end

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+        do
+                if [[ "Started" == $(volinfo_field "$volname" 'statusStr') ]];
```
nixpanic: Neither do you need the `;` here. It is only required if you place the `then` on the same line as the `if`.

In case the `$(...)` command fails, the string might be empty and the script can exit with an error. Place the `$(...)` between quotes.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+                                        then
+                                                wait_info="$vol_down_info"
+                                        else
+                                                wait_info="$wait_info, $vol_down_info"
+                                        fi
+                                fi
+                        fi
+                fi
+        done < <(gluster volume list)
+        if [[ ! -z "$wait_info" ]]; then echo "$wait_info"; fi
```
nixpanic: See, here the `;` is required.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function check_glusterd_running()
+{
+        if ! pidof glusterd > /dev/null 2>&1 ; then
```
nixpanic: Ooh, the horrors of inconsistently formatted `if` structures. It would be good to stick to a single format...

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+                exit 1;
+        fi
+}
+
+function wait_for_bricks_up()
+{
+        local endtime=$1
+        local now
+
+        while [[ ! -z "$(volumes_waiting_for_bricks_up)" ]]; do
```
nixpanic: Same as with if/then, decide if while/do should be on the same line or not, and stick to is for the whole script.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+                now=$(date +%s%N)
+                if [[ $now -ge $endtime ]]; then
```
nixpanic: Variables without qoutes, please add them.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+
+check_glusterd_running;
+timeout=$1
+start=$(date +%s%N)
+#Convert timeout to Nano Seconds as 'start' is in Nano seconds
+endtime=$(( timeout*1000000000 + start ))
+wait_for_bricks_up $endtime
+now=$(date +%s%N)
+printLog "INFO: All bricks for Block Hosting Volumes came online in $(((now-start)/1000000000)) seconds"
+exit 0;
```
nixpanic: No need for trailing `;` at the end of a line.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,156 @@
...
+        local vol_down_info
+        while read -r volname;
+        do
+                if [[ "Started" == $(volinfo_field "$volname" 'statusStr') ]];
+                then
+                        if [[ "Y" == "$(is_enabled "$(volinfo_option "$volname" features.shard)")" ]];
+                        then
+                                brick_count=$(volume_online_brick_count "$volname")
+                                total_brick_count=$(volinfo_field "$volname" 'brickCount')
+                                if [[ $brick_count -ne $total_brick_count ]];
```
nixpanic: More variables without `"${quotes}"`.

pranithk: Addressed the rest except this one as quotes are to handle values with spaces and these are numbers...

nixpanic: Ah, good point. You have to be extra careful with numbers. I've doublechecked the usage here, and I think it is safe without additional `-n "${brick_count}"`.

extras/wait-for-bricks.sh
```diff
@@ -0,0 +1,160 @@
...
+#later), or the GNU General Public License, version 2 (GPLv2), in all
+#cases as published by the Free Software Foundation.
+
+# This file waits for gluster block hosting volume bricks to be up. If all the
+# bricks are up, it exits with success status, otherwise waits for the $timeout
+# seconds before erroring out with the status about
+# online-bricks/expected-online-bricks
+
+LOGDIR="${GB_LOGDIR:-/var/log/gluster-block}"
+
```
pkalever: * Since this script will be run before one of gluster-blockd and tcmu-runner start, where the logdir might be absent, hence this script should ' # mkdir -p ${LOGDIR}' before it start anything.

* As we are anyway have to post a new revision, can you add one a line after line 22 

* may be you might have to miss to address my previous comment about trace output ?

* Also will it be possible to move the test script under '${srcdir}/tests/glusterfs/' ?

* last but not the least, please add tested-by/reviewed-by tags as part of the next revision.

Thanks @pranithk !

pkalever:
Thanks Pranith!

PR #108, opened 2018-08-06
wait-for-bricks: Use ps instead of pidof, improve performance

pranithk:
1) 'pidof' command may not exist in all setups, so changing that
    to use 'ps' command which is available more often. If 'ps' command is
    not found in the setup it prints that the command is not available
    in the setup in the error log.
    
    2) For finding if 10 block volumes are up and running it was taking 9 seconds
    when all the bricks are up. I changed the script to reduce the number of
    gluster commands that are executed to get this time to 1 seconds.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
 ACLOCAL_AMFLAGS = -I m4
 
-SUBDIRS = rpc utils cli daemon systemd docs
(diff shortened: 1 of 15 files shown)


pranithk:
Let me redo this again. I think I made a mistake

PR #109, opened 2018-08-06
wait-for-bricks: Use ps instead of pidof, improve performance

pranithk:
1) 'pidof' command may not exist in all setups, so changing that
to use 'ps' command which is available more often. If 'ps' command is
not found in the setup it prints that the command is not available
in the setup in the error log.

2) For finding if 10 block volumes are up and running it was taking 9 seconds
when all the bricks are up. I changed the script to reduce the number of
gluster commands that are executed to get this time to 1 seconds.

Signed-off-by: Pranith Kumar K <pkarampu@redhat.com>

### DIFF
--- a/extras/wait-for-bricks.sh
+++ b/extras/wait-for-bricks.sh
@@ -22,68 +22,24 @@ function printLog()
     echo "[$(date -u +'%Y-%m-%d %I:%M:%S')] ${msg}" >> "${LOGDIR}/gluster-block-bricks-start.log"
 }
 
-function volinfo_field()
-{
-    local vol=$1;
-    local field=$2;
-
-    gluster volume info "$vol" --xml | grep -oPm1 "(?<=<$field>)[^<]+"
-}
-
-function is_enabled()
-{
-        case "$1" in
-        0|off|no|false|disable)
-                echo "N"
-                ;;
-        1|on|yes|true|enable)
-                echo "Y"
-                ;;
-        *)
-                echo "N/A"
-                ;;
-        esac
-}
-
-function volinfo_option()
-{
-    local vol=$1;
-    local key=$2;
-
-    gluster volume get "$vol" "$key"| awk '{print $NF}' | tail -1
-}
-
-function check_brick_status() {
-        local volname=$1
-        local daemon=$2
-
-        cmd="gluster --xml volume status $volname"
-        if [[ -z "$daemon" ]]
-        then
-                 $cmd | grep -c '<status>1'
-        else
-                 $cmd | sed -n -e "/${daemon}/,/<status>/ p" | grep -c '<status>1'
-        fi
-}
-
 function volume_online_brick_count()
 {
         local volname=$1
-        local v1=0
-        local v2=0
-        local v3=0
-        local v4=0
-        local v5=0
-        local tot=0
+        local brick_count=$2
+        gluster --xml volume status "$volname" | grep '<status>' | head -n "$brick_count" | grep -c '<status>1'
+}
 
-        #First count total Number of bricks and then subtract daemon status
-        v1=$(check_brick_status "$volname")
-        v2=$(check_brick_status "$volname" "Self-heal")
-        v3=$(check_brick_status "$volname" "Quota")
-        v4=$(check_brick_status "$volname" "Snapshot")
-        v5=$(check_brick_status "$volname" "Tier")
-        tot=$((v1-v2-v3-v4-v5))
-        echo $tot
+function get_block_volnames_brick_count()
+{
+        gluster v info | \
+        # get volume name, status, shard-status, number-of-bricks
+        grep -P "(^Volume Name:|^Status:|^features.shard:|^Number of Bricks:)" | \
+        # get the 'started' volumes
+        sed '/^Volume Name:/{x;p;x;}' | sed -e '/./{H;$!d;}' -e 'x;/Status: Started/!d;' | \
+        # get shard enabled volumes
+        sed -e '/./{H;$!d;}' -e 'x;/features.shard: on/b' -e '/features.shard: yes/b' -e '/features.shard: enable/b' -e '/features.shard: true/b' -e '/features.shard: 1/b' -e d | \
+        # Print volume names
+        grep -P "(^Volume Name:|^Number of Bricks:)" | awk '{print $NF}'
 }
 
 function volumes_waiting_for_bricks_up()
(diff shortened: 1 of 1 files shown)


pranithk:
@pkalever @lxbsz please review these changes.

extras/wait-for-bricks.sh
```diff
@@ -92,31 +48,33 @@ function volumes_waiting_for_bricks_up()
...
+                printLog "ERROR: 'ps' command is not found"
+                exit 1;
+        fi
+
+# Explanation of why ps command is used this way:
+# https://stackoverflow.com/questions/9117507/linux-unix-command-to-determine-if-process-is-running
+        if ! ps cax | grep -w '[g]lusterd' > /dev/null 2>&1
         then
                 printLog "ERROR: Glusterd is not running";
                 exit 1;
```
lxbsz: BTW, the ';' is needed here ? I find some lines have it and some not :-)

extras/wait-for-bricks.sh
```diff
@@ -92,31 +48,33 @@ function volumes_waiting_for_bricks_up()
...
         if [[ ! -z "$wait_info" ]]; then echo "$wait_info"; fi
 }
 
 function check_glusterd_running()
 {
-        if ! pidof glusterd > /dev/null 2>&1
+        if ! command -v ps > /dev/null 2>&1
+        then
+                printLog "ERROR: 'ps' command is not found"
+                exit 1;
```
lxbsz: BTW, the ';' is needed here ? I find some lines have it and some not :-)

extras/wait-for-bricks.sh
```diff
@@ -92,34 +48,36 @@ function volumes_waiting_for_bricks_up()
...
 {
-        if ! pidof glusterd > /dev/null 2>&1
+        if ! command -v ps > /dev/null 2>&1
+        then
+                printLog "ERROR: 'ps' command is not found"
+                exit 1
+        fi
+
+# Explanation of why ps command is used this way:
+# https://stackoverflow.com/questions/9117507/linux-unix-command-to-determine-if-process-is-running
```
pkalever: Not exactly. This is a way tweaked version of what is been explained at stack-overflow, better remove it IMO.

extras/wait-for-bricks.sh
```diff
@@ -92,34 +48,36 @@ function volumes_waiting_for_bricks_up()
...
                 fi
-        done < <(gluster volume list)
+        done < <(block_volnames_brick_count_get)
         if [[ ! -z "$wait_info" ]]; then echo "$wait_info"; fi
 }
 
 function check_glusterd_running()
 {
-        if ! pidof glusterd > /dev/null 2>&1
+        if ! command -v ps > /dev/null 2>&1
```
pkalever: ps is part of package procps-ng, and should be defaultly shipped at all OS flavors.
IMO this check is not needed (just an opinion), and that should not block this patch :-)

extras/wait-for-bricks.sh
```diff
@@ -22,68 +22,24 @@ function printLog()
...
-        tot=$((v1-v2-v3-v4-v5))
-        echo $tot
+        gluster v info | \
+ # get volume name, status, shard-status, number-of-bricks
+        grep -P "(^Volume Name:|^Status:|^features.shard:|^Number of Bricks:)" | \
+ # get the 'started' volumes
+        sed '/^Volume Name:/{x;p;x;}' | sed -e '/./{H;$!d;}' -e 'x;/Status: Started/!d;' | \
+ # get shard enabled volumes
+        sed -e '/./{H;$!d;}' -e 'x;/features.shard: on/b' -e '/features.shard: yes/b' -e '/features.shard: enable/b' -e '/features.shard: true/b' -e '/features.shard: 1/b' -e d | \
+ #Print volume names
```
pkalever: If you are resubmitting this version, just add a space after \#

nixpanic: And indent the comment to be inline with the code.

extras/wait-for-bricks.sh
```diff
@@ -92,34 +48,36 @@ function volumes_waiting_for_bricks_up()
...
-        if ! pidof glusterd > /dev/null 2>&1
+        if ! command -v ps > /dev/null 2>&1
+        then
+                printLog "ERROR: 'ps' command is not found"
+                exit 1
+        fi
+
+# Explanation of why ps command is used this way:
+# https://stackoverflow.com/questions/9117507/linux-unix-command-to-determine-if-process-is-running
+        if ! ps cax | grep -w '[g]lusterd' > /dev/null 2>&1
```
nixpanic: instead of redirecting to `> /dev/null`, use `grep -q`

extras/wait-for-bricks.sh
```diff
@@ -92,34 +48,36 @@ function volumes_waiting_for_bricks_up()
...
         then
                 printLog "ERROR: Glusterd is not running";
-                exit 1;
+                exit 1
```
nixpanic: you removed the `;` after `exit`, that can be done at all end-of-lines

extras/wait-for-bricks.sh
pranithk: In containers, quite a few packages keep missing, for example, I have never seen an OS installtion without less, but I have seen quite a few containers without 'less' command. I don't think it hurts to keep this code. I don't mind adding the same checks for sed, grep, awk, head, echo as well.

extras/wait-for-bricks.sh
pranithk: Cool, will do

pranithk: Found this in manpage, so won't change this part:        Normally the exit status is 0 if a line is selected, 1 if no lines were selected, and
       2 if an error occurred.  However, if the -q or --quiet or --silent is used and a line
       is selected, the exit status is 0 even if an error occurred.

extras/wait-for-bricks.sh
pranithk: In one of the comments on that page:

> grep will also find itself running (e.g. ps cax | grep randomname will always return 0 because grep finds grep randomname (hope this is clear...). One fix is to add square brackets around the first letter of the process name, e.g. ps cax | grep [r]andomname. – Kyle G. Oct 29 '13 at 3:19

I think the only extra flag we added is -w for grep. I would like to keep it so that people who read this code would know why it is done this way.

extras/wait-for-bricks.sh
pranithk: Let me also restructure that part of the code.

nixpanic:
On Tue, Aug 07, 2018 at 01:41:24AM -0700, Pranith Kumar Karampuri wrote:
> pranithk commented on this pull request.
> 
> 
> 
> >          if [[ ! -z "$wait_info" ]]; then echo "$wait_info"; fi
>  }
>  
>  function check_glusterd_running()
>  {
> -        if ! pidof glusterd > /dev/null 2>&1
> +        if ! command -v ps > /dev/null 2>&1
> +        then
> +                printLog "ERROR: 'ps' command is not found"
> +                exit 1
> +        fi
> +
> +# Explanation of why ps command is used this way:
> +# https://stackoverflow.com/questions/9117507/linux-unix-command-to-determine-if-process-is-running
> +        if ! ps cax | grep -w '[g]lusterd' > /dev/null 2>&1
> 
> Found this in manpage, so won't change this part:        Normally the exit status is 0 if a line is selected, 1 if no lines were selected, and
>        2 if an error occurred.  However, if the -q or --quiet or --silent is used and a line
>        is selected, the exit status is 0 even if an error occurred.

It is very common to use 'grep -q' for not outputting any matching
lines. As long as the format (here '[g]lusterd') is correct and grep can
read from stdin (not passing a filename that may be nonexistent), it
should be safe. Are there any errors that you can think of which would
prevent using 'grep -q'?

pranithk:
> It is very common to use 'grep -q' for not outputting any matching lines. As long as the format (here '[g]lusterd') is correct and grep can read from stdin (not passing a filename that may be nonexistent), it should be safe.

Probably,

> Are there any errors that you can think of which would prevent using 'grep -q'?

I am not an expert on grep(I didn't even know grep -q existed ;-) ), but it is better to error out than continue when we face error in this scenario I feel.

pranithk:
Addressed everything else except 'grep -q' change.

extras/wait-for-bricks.sh
pkalever: I did not refer that part then. I refered our replace-node script for [] usage. Cool. I'm all okay

pkalever:
Tested this on the container env, with below cases:
* when glusterd down
* bricks down forever
* bricks down and brought up within timeout

All works for me.

These are my cosmic level suggestions, will leave this to your choice:
* s/block_volnames_brick_count_get/get_block_volnames_brick_count/
* s/check_/is_/
Ex: check_glusterd_running -> is_glusterd_running 

Tested-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>


Thanks @pranithk

pkalever:
@pranithk Sure.
Please update the PR and I will merge it.

Thanks!

pkalever:
Thanks @pranithk

PR #112, opened 2018-08-28
meta: reserve 10M space for meta-data store

pkalever:
Just like gluster file volume reserve 1% space, and Heketi reserves 2% space
currently. Reserving some space for block-meta within block hosting volume can
solve some potential issues.

Problem:
-------
Say, if the block hosting volume is 100% used, we expect the block volume delete
operation should free some space. But since the delete operation will also incur
some metadata writes into the respective block meta file, the meta writes will
fail with ENOSPC. Hence the block delete operation will be terminated.

Solution:
--------
On every create request we should do glfs_statvfs() on the block hosting volume,

```
if (($freeSize - $reserveSpace) >= reqBlockSize) {
   ---
   proceed with the block create request
   ---
} else {
   ---
   fail the block create request
   ---
}
```

How much reservation needs to be done ?

```
[root@localhost ~]# ls /mnt/block-meta/ -lash
total 13K
4.0K d--------- 2 root root 4.0K Aug 28 20:16 .
4.0K drwxr-xr-x 6 root root 4.0K Aug 28 12:06 ..
 512 -rw------- 1 root root  357 Aug 28 20:16 block1
 512 -rw------- 1 root root  356 Aug 28 20:13 block10
 512 -rw------- 1 root root  356 Aug 28 20:13 block2
 512 -rw------- 1 root root  356 Aug 28 20:13 block3
 512 -rw------- 1 root root  356 Aug 28 20:13 block4
 512 -rw------- 1 root root  357 Aug 28 20:13 block5
 512 -rw------- 1 root root  356 Aug 28 20:13 block6
 512 -rw------- 1 root root  356 Aug 28 20:13 block7
 512 -rw------- 1 root root  357 Aug 28 20:13 block8
 512 -rw------- 1 root root  356 Aug 28 20:13 block9
   0 -rw------- 1 root root    0 Aug 28 12:06 meta.lock
   0 -rw------- 1 root root    0 Aug 28 12:06 prio.info
[root@localhost ~]#
```

Note that each block file takes 512B and they can grow to 4K easily. Assuming
we can scale upto 1000 block volumes, so block metadata should take 4M max.
But rounding this to 10M should not hurt.

Thanks to Raghavendra Talur \<rtalur@redhat.com\> for contributing to the
idea.

### DIFF
--- a/rpc/glfs-operations.c
+++ b/rpc/glfs-operations.c
@@ -88,13 +88,50 @@ glusterBlockVolumeInit(char *volume, int *errCode, char **errMsg)
 
 
 int
+glusterBlockCheckAvailableSpace(struct glfs *glfs,
+                                char *volume, size_t blockSize, char **errMsg)
+{
+  struct statvfs buf = {'\0', };
+  int errSave = 0;
+
+
+  if (!glfs_statvfs(glfs, "/", &buf)) {
+    if ((buf.f_bfree * buf.f_bsize) >= GB_METASTORE_RESERVE + blockSize) {
+      return 0;
+    }
+    LOG("gfapi", GB_LOG_ERROR,
+        "glfs_statvfs('%s'): Low space on volume => "
+        "Total size: %lu, Free space: %lu, Block request space: %lu", volume,
+        buf.f_blocks * buf.f_bsize, buf.f_bfree * buf.f_bsize, blockSize);
+    GB_ASPRINTF(errMsg, "Low space on the volume %s\n", volume);
+    errSave = ENOSPC;
+  } else {
+    errSave = errno;
+    LOG("gfapi", GB_LOG_ERROR,
+        "glfs_statvfs('%s'): couldn't get file-system statistics", volume);
+    GB_ASPRINTF(errMsg,
+                "couldn't get file-system statistics on volume %s\n", volume);
+  }
+  errno = errSave;
+
+  return -1;
+}
+
+
+int
 glusterBlockCreateEntry(struct glfs *glfs, blockCreateCli *blk, char *gbid,
                         int *errCode, char **errMsg)
 {
   struct glfs_fd *tgfd;
   struct stat st;
   char *tmp;
-  int ret;
+  int ret = -1;
+
+
+  if (glusterBlockCheckAvailableSpace(glfs, blk->volume, blk->size, errMsg)) {
+    *errCode = errno;
+    goto out;
+  }
 
   ret = glfs_mkdir (glfs, GB_STOREDIR, 0);
   if (ret && errno != EEXIST) {
@@ -271,6 +308,12 @@ glusterBlockResizeEntry(struct glfs *glfs, blockModifySize *blk,
       goto close;
     }
 
+    if (glusterBlockCheckAvailableSpace(glfs, blk->volume, blk->size - sb.st_size, errMsg)) {
+      *errCode = errno;
+      ret = -1;
+      goto close;
+    }
+
     ret = glfs_ftruncate(tgfd, blk->size);
     if (ret) {
       *errCode = errno;
@@ -307,8 +350,10 @@ glusterBlockResizeEntry(struct glfs *glfs, blockModifySize *blk,
 
  out:
   if (ret) {
-    GB_ASPRINTF (errMsg, "Not able to resize storage for %s/%s [%s]",
-                 blk->volume, blk->block_name, strerror(*errCode));
+    if (errMsg && !(*errMsg)) {
+      GB_ASPRINTF (errMsg, "Not able to resize storage for %s/%s [%s]",
+                   blk->volume, blk->block_name, strerror(*errCode));
+    }
 
   }
 
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -52,8 +52,10 @@
 
 # define  SUN_PATH_MAX           (sizeof(struct sockaddr_un) - sizeof(unsigned short int)) /*sun_family*/
 
-# define GB_TIME_STRING_BUFLEN  \
-         (4 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 6 + 1 +   5)
+# define  GB_METASTORE_RESERVE   10485760   /* 10 MiB reserve for block-meta */
+
+# define  GB_TIME_STRING_BUFLEN  \
+          (4 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 6 + 1 +   5)
      /*   Yr      Mon     Day     Hour    Min     Sec     Ms  NULL  Round-off(32)
          2017 -   06  -   01  ' ' 18  :   58  :   29  . 695147 '\0' Power^2
          2017-06-01 18:58:29.695147                                             */


pkalever:
please-review: @pranithk @lxbsz @bhumikagoyal @obnoxxx @raghavendra-talur @humblec @amarts

humblec:
Lgtm. Thanks @pkalever !

pranithk:
@pkalever What is the idea when the block devices are sparse? At the time of create the size would be more than 10M but over time it can fill up the storage leading to same problem?

Another corner case to consider is when someone keeps executing resize in a loop or change auth in a loop, they can go over the size limit we think it may not cross. So even with this fix, we can hit the issue.

If you want to address those problems in a separate issue/PR I will go ahead and review this one, otherwise I will hold on until the next revision comes in.

pkalever:
> @pkalever What is the idea when the block devices are sparse? At the time of create the size would be more than 10M but over time it can fill up the storage leading to same problem?

This is a very good point, I was actually expecting this question :-)

Restricting the block create with sparse files committed/file-size rather than actual disk usage might be a less efficient way of using the BHV/underline-storage.

If we still want to solve this with sparse files as well at the cost of space efficiency, then I'm not sure if there is good way to get the committed size rather than actual disk usage ?

> Another corner case to consider is when someone keeps executing resize in a loop or change auth in a loop, they can go over the size limit we think it may not cross. So even with this fix, we can hit the issue.

Nope, we have added the check in resize function as well, so this should not be a problem. Let me know if you still see any corner case.

Thanks @pranithk !

lxbsz:
Test this and it works as expected:

1, the backend disk size is 2G prealloced, like:
[root@rhel3 gluster-block]# lsblk 
NAME          MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sdc             8:32   0    2G  0 disk /data-test

2, create one volume like:
[root@rhel3 gluster-block]# gluster volume create dht-test replica 3 rhel1:/data-test/rep1 rhel2:/data-test/rep2 rhel3:/data-test/rep3

3, create target devices:
[root@rhel3 gluster-block]# gluster-block create dht-test/block1 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 1G
IQN: iqn.2016-12.org.gluster-block:ec7d6167-c32e-49af-9796-3e137ea3a373
PORTAL(S):  192.168.195.157:3260 192.168.195.158:3260 192.168.195.159:3260
RESULT: SUCCESS
===========
[root@rhel3 gluster-block]# gluster-block create dht-test/block2 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 1G
Low space on the volume dht-test

RESULT:FAIL
==========
[root@rhel3 gluster-block]# gluster-block create dht-test/block2 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 512m
IQN: iqn.2016-12.org.gluster-block:aeb1e9a1-f2c3-4de2-8d05-4e8c4c5b79ce
PORTAL(S):  192.168.195.157:3260 192.168.195.158:3260 192.168.195.159:3260
RESULT: SUCCESS
==========
[root@rhel3 gluster-block]# df /mnt/
Filesystem     1K-blocks    Used Available Use% Mounted on
rhel1:dht-test   2086912 1627008    459904  78% /mnt
[root@rhel3 gluster-block]# df -h /mnt/
Filesystem      Size  Used Avail Use% Mounted on
rhel1:dht-test  2.0G  1.6G  450M  78% /mnt
==========
[root@rhel3 gluster-block]# gluster-block create dht-test/block3 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 440m
Low space on the volume dht-test

RESULT:FAIL
==========
[root@rhel3 gluster-block]# gluster-block create dht-test/block3 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 400m
IQN: iqn.2016-12.org.gluster-block:9112eaf7-4af9-4832-966c-6fa8c2da4b2e
PORTAL(S):  192.168.195.157:3260 192.168.195.158:3260 192.168.195.159:3260
RESULT: SUCCESS
==========
[root@rhel3 gluster-block]# df -h /mnt/
Filesystem      Size  Used Avail Use% Mounted on
rhel1:dht-test  2.0G  2.0G   50M  98% /mnt
==========
[root@rhel3 gluster-block]# gluster-block create dht-test/block4 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 40m
Low space on the volume dht-test

RESULT:FAIL
==========
[root@rhel3 gluster-block]# gluster-block create dht-test/block4 ha 3 prealloc full 192.168.195.157,192.168.195.158,192.168.195.159 39m
IQN: iqn.2016-12.org.gluster-block:17f9a25d-a075-4a71-a209-65696b3ca269
PORTAL(S):  192.168.195.157:3260 192.168.195.158:3260 192.168.195.159:3260
RESULT: SUCCESS

And the same test for resize.

Tested-by: Xiubo Li <xiubli@redhat.com>

utils/utils.h
```diff
@@ -52,8 +52,10 @@
 
 # define  SUN_PATH_MAX           (sizeof(struct sockaddr_un) - sizeof(unsigned short int)) /*sun_family*/
 
-# define GB_TIME_STRING_BUFLEN  \
-         (4 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 6 + 1 +   5)
+# define  GB_METASTORE_RESERVE   10485760   /* 10 MiB reserve for block-meta */
```
obnoxxx: Can we make this configurable, defaulting to 10 MiB?

rpc/glfs-operations.c
```diff
@@ -307,8 +350,10 @@ glusterBlockResizeEntry(struct glfs *glfs, blockModifySize *blk,
 
  out:
   if (ret) {
-    GB_ASPRINTF (errMsg, "Not able to resize storage for %s/%s [%s]",
-                 blk->volume, blk->block_name, strerror(*errCode));
+    if (errMsg && !(*errMsg)) {
```
obnoxxx: Is this sneaking in an independent change that should be a different patch/fix?
Or else please explain how this is related to the 10MiB space reservation change.

utils/utils.h
pkalever: I do not find this useful ATM. And do we have a case for this where we might want to alter reserved space ?

Does Heketi  provide an option to tune reserved space ?

rpc/glfs-operations.c
pkalever: Oh yes! In absence of this, the errMsg string that we filled at glusterBlockCheckAvailableSpace() see line 311, will be overridden.

pranithk:
> Restricting the block create with sparse files committed/file-size rather than actual disk usage might be a less efficient way of using the BHV/underline-storage.
> If we still want to solve this with sparse files as well at the cost of space efficiency, then I'm not sure if there is good way to get the committed size rather than actual disk usage ?

I think the way to solve this problem is to make sure the meta files don't run out of space.
One way to achieve it is, we should change the current way of doing transaction logging in such a way that the size of the meta file won't be more than 'X' bytes and when we initially create the file we should fallocate for that size at the time of creating the meta file.

> Nope, we have added the check in resize function as well, so this should not be a problem. Let me know if you still see any corner case.

I am talking about the growth in meta file if we keep executing the resize/auth command again and again

utils/utils.h
obnoxxx: touche! :-)
On the other hand heketi uses a percentage of the raw capacity instead of a constant size :-)

rpc/glfs-operations.c
obnoxxx: Oh, that treatment of errMsg is just plain horrible! :-)
But I get it. That complaint is not for this patch then ... ;-)

pkalever:
> I think the way to solve this problem is to make sure the meta files don't run out of space.
One way to achieve it is, we should change the current way of doing transaction logging in such a way that the size of the meta file won't be more than 'X' bytes and when we initially create the file we should fallocate for that size at the time of creating the meta file.

Yes, this is also a good solution for this bug.

> I am talking about the growth in meta file if we keep executing the resize/auth command again and again

In current model, we took 10K per file, since we may scale to 1000 now, we just reserved 10M.
Even if we fallocate the metafile with 10K, repeated resize may exceed this limit too.

In this case:

Advantage & disadvantages:
-------

* With current model:
10M space will be shared across the metafile (1 meta file can be 512B and other can be 10K or even more), but with sparse files this won't work well.

* With fallocate model:
This will address issue with sparse block backend files, but If individual metadata file size reached 10K (fallocate size) then we might face issues.

Let me know what you prefer.

Thanks @pranithk !

pranithk:
> we should change the current way of doing transaction logging in such a way that the size of the meta file won't be more than 'X' bytes and when we initially create the file we should fallocate for that size at the time of creating the meta file.

I think you missed this point while coming up with advantages & disadvantages. Our logging model has to change so that the size of the metafile won't exceed a certain limit. One way to achieve that is by only storing the final configuration. When a new transaction happens, we do whatever we do at the moment on the metafile with relevant states, but at the end when the transaction completes, we store the final configuration.

pranithk:
> I think you missed this point while coming up with advantages & disadvantages. Our logging model has to change so that the size of the metafile won't exceed a certain limit. One way to achieve that is by only storing the final configuration. When a new transaction happens, we do whatever we do at the moment on the metafile with relevant states, but at the end when the transaction completes, we store the final configuration.

I think it is better to split the journaling and configuration into separate files to make the implementation easier. We can spend more time on the design based on what is the future direction of the configuration store. If there is an integration plan with GD2, I think configuration will be stored in etcd, so we don't have to probably spend time doing this and probably can live with whatever you sent. So let me know what the plan is.

pkalever:
All good points @pranithk!

Since we do not have a good way to decide whats needs to be done after the metafile reached 10K (fallocate size per meta), and for this model to work smoothly, storing final cut of effective configuration will work very well.

I understand the significance of tweaking our current way of storing meta info. And yes, we need to pick this very soon.

To solve the problem at usecases that purely use prealloc files, such as cns, I think pulling this PR-in, will help, rather than waiting for metadata reconstruction along with fixed-size meta-files work.

I will wait for your ack.

Thanks!

pranithk:
@pkalever sure go ahead with this approach for now.

pkalever:
Just updated the commit msg with reviewby and testby tags.

Issue #117, opened 2018-09-06
Dump all failure cli responses to stderr


pkalever:
On FAILURE dump reply.out to stderr
On SUCCESS dump reply.out to stdout

pranithk:
@pkalever Traditionally applications dump to stderr on Failure and stdout on success, wondering what prompted you to dump to stderr even on success.

pranithk:
@pkalever Didn't get you. Are you saying On SUCCESS we dump it on stdout? Issue description says: " On SUCCESS dump reply.out to stderr"

pkalever:
> @pkalever Didn't get you. Are you saying On SUCCESS we dump it on stdout? Issue description says: " On SUCCESS dump reply.out to stderr"

Ah! sorry. Just corrected. Thanks!

PR #121, opened 2018-09-25
gluster-block: dump all failure msgs to stderr

pkalever:
Ensure:
On FAILURE dump error msg to stderr
On SUCCESS dump response to stdout

Fixes: #117

### DIFF
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -121,15 +121,11 @@
             pthread_mutex_unlock(&x);                                \
           } while (0)
 
-# define ERROR(fmt, ...)                                             \
-         do {                                                        \
-           fprintf(stderr, "Error: " fmt " [at %s+%d :<%s>]\n",      \
-                   __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);   \
-          } while (0)
-
-# define  MSG(fmt, ...)                                              \
+# define  MSG(fd, fmt, ...)                                          \
           do {                                                       \
-            fprintf(stdout, fmt, __VA_ARGS__);                       \
+            if (fd <= 0)       /* including STDIN_FILENO 0 */        \
+              fd = stderr;                                           \
+            fprintf(fd, fmt, __VA_ARGS__);                           \
           } while (0)
 
 
--- a/cli/gluster-block.c
+++ b/cli/gluster-block.c
@@ -35,7 +35,7 @@
 # define  GB_ARGCHECK_OR_RETURN(argcount, count, cmd, helpstr)        \
           do {                                                        \
             if (argcount != count) {                                  \
-              MSG("Inadequate arguments for %s:\n%s\n", cmd, helpstr);\
+              MSG(stderr, "Inadequate arguments for %s:\n%s\n", cmd, helpstr); \
               return -1;                                              \
             }                                                         \
           } while(0)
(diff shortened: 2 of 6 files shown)


pkalever:
@lxbsz please review.

Thanks!

utils/utils.h
```diff
@@ -121,15 +121,11 @@
...
-           fprintf(stderr, "Error: " fmt " [at %s+%d :<%s>]\n",      \
-                   __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);   \
-          } while (0)
-
-# define  MSG(fmt, ...)                                              \
+# define  MSG(fd, fmt, ...)                                          \
           do {                                                       \
-            fprintf(stdout, fmt, __VA_ARGS__);                       \
+            if (fd < 0)                                              \
+              fd = stderr;                                           \
```
lxbsz: Here I think it should be fd > 0, the 0 is the stdin.
Or here it should only allow stdout and stderr ?

utils/utils.h
```diff
@@ -121,15 +121,11 @@
...
+            fprintf(fd, fmt, __VA_ARGS__);                           \
```
lxbsz: Here how about adding the __FILE__, __LINE__ and __FUNCTION__ in the log messages ?

utils/utils.h
pkalever: > Here I think it should be fd > 0

No

Usually fd < 0 are not valid, so route the output to stderr (fd = stderr)

If stderr, is supplied the output __VA_ARGS__ should be printed on stderr
if stdout, is supplied the output should be printed on stdout.
If fd < 0  is supplied, lets print the output to stderr.

For reference
```
STDIN_FILENO  
Standard input value stdin. Its value is 0.

STDOUT_FILENO  
Standard output value stdout. Its value is 1.

STDERR_FILENO  
Standard error value stderr. Its value is 2.
```

Did I miss something here ?

utils/utils.h
pkalever: No these are dumped to console. For example, the create cli response output:
```
# gluster-block create block-test/sample-block ha 3 192.168.1.11,192.168.1.12,192.168.1.13 1GiB
IQN: iqn.2016-12.org.gluster-block:aafea465-9167-4880-b37c-2c36db8562ea
PORTAL(S): 192.168.1.11:3260 192.168.1.12:3260 192.168.1.13:3260
RESULT: SUCCESS
````

It doesn't make any sense to print the FILE, LINE and FUNCTION  in this case.

For error outputs, it might makesense in some scenarios, but we will get these info in the LOG's files.

**Comment by pkalever** [comment]
@lxbsz thanks for the review. Please checkout my responses.

File: utils/utils.h (hunk 2 above)
Comment (lxbsz) [reply utils/utils.h]: yeah, that's good.

File: utils/utils.h (hunk 1 above)
Comment (lxbsz) [reply utils/utils.h]: I just wanted to make sure that you were aware that when the fd == 0, which is the stdin. Will this make sense here ? I am okay with this.

Comment (pkalever) [reply utils/utils.h]: Ah! so you are saying line 126 should rather be 
```
if (fd <= 0) {
  fd = stderr;
}
```

I'm okay with this too. Let me know, if this is what you have request for.

Thanks @lxbsz

lxbsz: Yeah, this was what I meant :)

pkalever: Cool :-)
Will refresh the PR with this change very soon.

Thanks @lxbsz

pkalever:
@lxbsz updated with the requested change. Thanks!

lxbsz:
Looks good to me.

Reviewed-by: Xiubo Li <xiubli@redhat.com>

pkalever:
@amarts is this PR ready to dive in master ?

utils/utils.h
amarts: This mostly needs taking care? Or is it already done!

pkalever: @amarts What are we looking for ?
According to us, we are done with this.

amarts: From the comment, I didn't realize if it was done! Hence the comment. Patch looks OK for me.

pkalever: Cool.

pkalever:
@amarts @lxbsz Thanks for the review. Merging the PR now.

PR #122, opened 2018-09-28
spec: create and package /var/log/gluster-block/

nixpanic:
Usually, the daemon creates the log directory.
But when it fails to start, and the log directory
has not been pre-created, then the client will
segfault instead of logging useful errors.

Change-Id: I55b5c9fcd2a9a97dba47e87a826606d48f8ac8bd
Signed-off-by: Michael Adam <obnox@redhat.com>
Signed-off-by: Niels de Vos <ndevos@redhat.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,6 +29,9 @@ srpm: prep
 prep: dist
 	@mkdir -p build/rpmbuild/{BUILD,SPECS,RPMS,SRPMS,SOURCES}
 
+install-data-local:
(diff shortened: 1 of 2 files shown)


nixpanic:
The original change came from @obnoxxx via https://review.gluster.org/c/gluster-block/+/17613
My modification has been to create the directory in the `Makefile.am`.

pkalever:
Had some conflicts with master, had to manually resolve and merge them through cli.

Thanks @nixpanic!

PR #128, opened 2018-10-10
genconfig: with old blocks use first valid ipaddr from metadata as PR…

pkalever:
…IOPATH

for block volumes which doesn't have PRIOPATH set (created with older -
versions, before introducing load-balance RFE), lets use first valid ip address
from metadata file as prio_path.

This will help mainatin compatibility of old block volumes with 'prio "alua"'
multipath conf recommondation.


pkalever:
@lxbsz @amarts please review. Thanks!

pkalever:
Updated the tags. Thanks.

PR #135, opened 2018-10-18
CI: add .travis.yml for testing

amarts:
[WIP] Just for testing, good to go after tests.

Signed-off-by: Amar Tumballi <amarts@redhat.com>


obnoxxx:
This fails because the glusterfs-4.1 ppa does not provide trusty packages.

Issue #136, opened 2018-10-18
Provide an option to fetch the volume info from different place, other than 'localhost'.


amarts:
In `glfs-operations.c` we have below line:

```
rpc/glfs-operations.c:  ret = glfs_set_volfile_server(glfs, "tcp", "localhost", 24007);
```

For keeping the sanity, we should 'default' to localhost:24007, but user should have an option to fetch the details for the Block Hosting volume from different IP (or hostname). That would allow us to separately host gluster-block from glusterfs bricks, where there is glusterd running.

pkalever:
Fixed-by: https://github.com/gluster/gluster-block/pull/161

PR #137, opened 2018-10-18
arpg: provide an option to set glusterd hostname and port

amarts:
Fixes: #136 
Signed-off-by: Amar Tumballi <amarts@redhat.com>

### DIFF
--- a/daemon/gluster-blockd.c
+++ b/daemon/gluster-blockd.c
@@ -46,6 +46,8 @@ glusterBlockDHelp(void)
       "  gluster-blockd [--glfs-lru-count <COUNT>]\n"
       "                 [--log-level <LOGLEVEL>]\n"
       "                 [--no-remote-rpc]\n"
+      "                 [--glusterd-hostname]\n"
+      "                 [--glusterd-port]\n"
       "\n"
       "commands:\n"
       "  --glfs-lru-count <COUNT>\n"
@@ -56,6 +58,12 @@ glusterBlockDHelp(void)
       "  --no-remote-rpc\n"
       "        Ignore remote rpc communication, capabilities check and\n"
       "        other node sanity checks\n"
+      "  --glusterd-hostname\n"
+      "        hostname of where to fetch the volume details for block\n"
+      "        hosting volume [Default: localhost]\n"
+      "  --glusterd-port\n"
+      "        port of glusterd, either on localhost or remote.\n"
+      "        [Default: 24007]\n"
       "  --help\n"
       "        Show this message and exit.\n"
       "  --version\n"
@@ -266,6 +274,28 @@ glusterBlockDParseArgs(int count, char **options)
       gbConf.noRemoteRpc = true;
       break;
 
+    case GB_DAEMON_GLUSTERD_HOST:
+      if (count - optind  < 1) {
+        MSG(stderr, "option '%s' needs argument <hostname>\n",
+            options[optind-1]);
+        return -1;
+      }
+      if (GB_STRDUP(gbConf.glusterdHostname, options[optind]) < 0) {
+        MSG(stderr, "failed to copy hostname: %s\n", options[optind]);
+        return -1;
+      }
+      break;
+
+    case GB_DAEMON_GLUSTERD_PORT:
+      if (count - optind  < 1) {
+        MSG(stderr, "option '%s' needs argument <port>\n", options[optind-1]);
+        return -1;
+      }
+      if (sscanf(options[optind], "%d", &gbConf.glusterdPort) != 1) {
+        MSG(stderr, "failed to get port details: %s\n", options[optind]);
+        return -1;
+      }
+      break;
     }
 
     optind++;
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -4201,6 +4201,7 @@ block_create_common(blockCreate *blk, char *rbsize, char *prio_path)
   char *authcred = NULL;
   char *save = NULL;
   char *exec = NULL;
+  char *gdHostname = NULL;
   blockResponse *reply = NULL;
   blockServerDefPtr list = NULL;
   size_t i;
@@ -4221,9 +4222,15 @@ block_create_common(blockCreate *blk, char *rbsize, char *prio_path)
     prioCap = true;
   }
 
+  /* If it is known that glusterd is hosted in a
+     given hostname, use that instead */
+  gdHostname = gbConf.glusterdHostname;
+  if (!strcmp(gdHostname, "localhost")) {
+    gdHostname = blk->ipaddr;
+  }
   if (GB_ASPRINTF(&backstore, "%s %s name=%s size=%zu cfgstring=%s@%s%s/%s%s wwn=%s",
                   GB_TGCLI_GLFS_PATH, GB_CREATE, blk->block_name, blk->size,
-                  blk->volume, blk->ipaddr, GB_STOREDIR, blk->gbid,
+                  blk->volume, gdHostname, GB_STOREDIR, blk->gbid,
                   rbsize ? rbsize: "", blk->gbid) == -1) {
     goto out;
   }
--- a/rpc/glfs-operations.c
+++ b/rpc/glfs-operations.c
@@ -36,7 +36,8 @@ glusterBlockVolumeInit(char *volume, int *errCode, char **errMsg)
     return NULL;
   }
 
-  ret = glfs_set_volfile_server(glfs, "tcp", "localhost", 24007);
+  ret = glfs_set_volfile_server(glfs, "tcp", gbConf.glusterdHostname,
+                                gbConf.glusterdPort);
   if (ret) {
     *errCode = errno;
     GB_ASPRINTF (errMsg, "Not able to add Volfile server for volume %s[%s]",
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -19,7 +19,9 @@
 struct gbConf gbConf = {
   .glfsLruCount = LRU_COUNT_DEF,
   .logLevel = GB_LOG_INFO,
-  .logDir = GB_LOGDIR
+  .logDir = GB_LOGDIR,
+  .glusterdHostname = "localhost",
+  .glusterdPort = 24007,
 };
 
 const char *argp_program_version = ""                                 \
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -140,6 +140,10 @@ struct gbConf {
   pthread_mutex_t lock;
   char cmdhistoryLogFile[PATH_MAX];
   bool noRemoteRpc;
+
+  /* Required only if blockhosting volume is in a remote node */
+  char *glusterdHostname;
+  int glusterdPort;
 };
 
 extern struct gbConf gbConf;
@@ -425,6 +429,8 @@ typedef enum gbDaemonCmdlineOption {
   GB_DAEMON_GLFS_LRU_COUNT = 4,
   GB_DAEMON_LOG_LEVEL      = 5,
   GB_DAEMON_NO_REMOTE_RPC  = 6,
+  GB_DAEMON_GLUSTERD_HOST  = 7,
+  GB_DAEMON_GLUSTERD_PORT  = 8,
 
   GB_DAEMON_OPT_MAX
 } gbDaemonCmdlineOption;
@@ -437,6 +443,8 @@ static const char *const gbDaemonCmdlineOptLookup[] = {
   [GB_DAEMON_GLFS_LRU_COUNT] = "glfs-lru-count",
   [GB_DAEMON_LOG_LEVEL]      = "log-level",
   [GB_DAEMON_NO_REMOTE_RPC]  = "no-remote-rpc",
+  [GB_DAEMON_GLUSTERD_HOST]  = "glusterd-hostname",
+  [GB_DAEMON_GLUSTERD_PORT]  = "glusterd-port",
 
   [GB_DAEMON_OPT_MAX]        = NULL,
 };


amarts:
@vbellur @lxbsz @pranithk @pkalever

daemon/gluster-blockd.c
```diff
@@ -112,11 +120,11 @@ glusterBlockCliThreadProc (void *vargp)
 
   if (!svc_register(transp, GLUSTER_BLOCK_CLI, GLUSTER_BLOCK_CLI_VERS,
                     gluster_block_cli_1, IPPROTO_IP)) {
-		LOG("mgmt", GB_LOG_ERROR,
+                LOG("mgmt", GB_LOG_ERROR,
         "unable to register (GLUSTER_BLOCK_CLI, GLUSTER_BLOCK_CLI_VERS: %s)",
         strerror (errno));
     goto out;
-	}
+        }
```
lxbsz: There is nothing changed except adding some spaces.

daemon/gluster-blockd.c
```diff
@@ -266,6 +274,29 @@ glusterBlockDParseArgs(int count, char **options)
       gbConf.noRemoteRpc = true;
       break;
 
+    case GB_DAEMON_GLUSTERD_HOST:
+      if (count - optind  < 1) {
+        MSG("option '%s' needs argument <hostname>\n", options[optind-1]);
+        return -1;
+      }
+      if (GB_STRDUP(gbConf.glusterdHostname, options[optind]) < 0) {
+        LOG("mgmt", GB_LOG_ERROR, "failed to copy hostname: %s\n",
```
lxbsz: Should we use MSG instead here ?

daemon/gluster-blockd.c
```diff
@@ -266,6 +274,29 @@ glusterBlockDParseArgs(int count, char **options)
...
+
+      break;
+
+    case GB_DAEMON_GLUSTERD_PORT:
+      if (count - optind  < 1) {
+        MSG("option '%s' needs argument <port>\n", options[optind-1]);
+        return -1;
+      }
+      if (sscanf(options[optind], "%d", &gbConf.glusterdPort) != 1) {
+        LOG("mgmt", GB_LOG_ERROR, "failed to get port details: %s\n",
```
lxbsz: And here ?

daemon/gluster-blockd.c
```diff
@@ -266,6 +274,29 @@ glusterBlockDParseArgs(int count, char **options)
...
+      if (count - optind  < 1) {
+        MSG("option '%s' needs argument <hostname>\n", options[optind-1]);
+        return -1;
+      }
+      if (GB_STRDUP(gbConf.glusterdHostname, options[optind]) < 0) {
+        LOG("mgmt", GB_LOG_ERROR, "failed to copy hostname: %s\n",
+            options[optind]);
+        return -1;
+      }
+
```
lxbsz: Useless space line.

rpc/glfs-operations.c
```diff
@@ -36,7 +38,12 @@ glusterBlockVolumeInit(char *volume, int *errCode, char **errMsg)
     return NULL;
   }
 
-  ret = glfs_set_volfile_server(glfs, "tcp", "localhost", 24007);
+  port = gbConf.glusterdPort;
+  hostname = gbConf.glusterdHostname;
+
+  ret = glfs_set_volfile_server(glfs, "tcp",
+                                ((hostname)? hostname: "localhost"),
+                                ((port)? port: 24007));
```
lxbsz: Are the "()" needed for hostname and port here ?

lxbsz: Maybe we should init the default value in:
`
20 struct gbConf gbConf = {LRU_COUNT_DEF, GB_LOG_INFO, GB_LOGDIR, '\0', '\0', '\0', '\0', '\0'};
`

utils/utils.c
```diff
@@ -17,7 +17,7 @@
 # include "config.h"
 
 
-struct gbConf gbConf = {LRU_COUNT_DEF, GB_LOG_INFO, GB_LOGDIR, '\0', '\0', '\0', '\0', '\0'};
+struct gbConf gbConf = {LRU_COUNT_DEF, GB_LOG_INFO, GB_LOGDIR, '\0', '\0', '\0', '\0', '\0', '\0', 0};
```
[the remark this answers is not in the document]
amarts: I added last entry as 'int port', and hence it makes sense to keep it as 0, and not '\0', even though technically both are same value, the type would differ.

Anyways, fixed this initialization as part of #141 later..

amarts:
@lxbsz addressed all the comments

amarts:
@lxbsz I saw your comments (through email, not here). I like the idea! But if you and @pkalever are OK with #141 PR, i will rebase this patch on top of that!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_set_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_set_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

pkalever:
@amarts @lxbsz @obnoxxx 

There are certain design challenges here, I remember @pranithk, @vbellur and myself had discussions  about this initially.

1. glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

2. glusterd hostname and port should be clubbed together with a delimiter as a separator,  to reduce the args in cli, something like "192.169.0.1:24007"

3. did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

4. right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - ***lets work on this first -  I will get back to you on this asap***

5. Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_**get**_volfile_server(glfs); 

Thanks!

amarts:
> glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.

This patch tries to achieve it.

> glusterd hostname and port should be clubbed together with a delimiter as a separator, to reduce the args in cli, something like "192.169.0.1:24007"

Lets not bother for now about this, and keep them as separate as number of argument is not an issue of concern?

> did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.

Let me give perspective from GCS, where it matters. There will always be a service running, so you will have a hostname, which will resolve and get you a volfile.

> right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - lets work on this first - I will get back to you on this asap
Not an immediate blocker, IMO! As we are looking to deploy it in cloud, there will always be a service, lets solve this later.

> Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_get_volfile_server(glfs);

Because the above is not a requirement, I don't think we should be bothered about this!

lxbsz:
On 2018/10/22 13:07, Amar Tumballi wrote:
>
> @lxbsz <https://github.com/lxbsz> I saw your comments (through email, 
> not here). I like the idea! But if you and @pkalever 
> <https://github.com/pkalever> are OK with #141 
> <https://github.com/gluster/gluster-block/pull/141> PR, i will rebase 
> this patch on top of that!
>
Yeah, Amar, I am okay with that.

Thanks.

pkalever:
> > glusterd-hostname should be supplied to gluster-block create cli (per block volume). This gives us flexibility to have block hosting volumes in different clusters.
> 
> This patch tries to achieve it.

@amarts 

I see changes to daemon and not to the cli process as part of this PR. why do we want the volserver (glusterd-hostname) option tied to daemon process.

(BHV -> Block hosting volume)
Say if we have a BHV1 (IP1, IP2, IP3) and BHV2 (IP4, IP5, IP6) both in different clusters, how you are going to supply the IP's to the dameon now ?

assuming , you will fix that by csv format,
Post the daemon startup, user wants to use BHV3 (part of different cluster, IP7, IP8, IP9), how this solution works ? Daemon-restarts are not recommended. 

gluster-blockd is started by systemd unit today, and how are we going to supply these IP's to the dameon during start time ? Did we thought about this ? even if we want to supply IP's at /etc/sysconfig/gluster-blockd, then top layers will have difficulties automating this.

> 
> > glusterd hostname and port should be clubbed together with a delimiter as a separator, to reduce the args in cli, something like "192.169.0.1:24007"
> 
> Lets not bother for now about this, and keep them as separate as number of argument is not an issue of concern?

See, https://github.com/gluster/gluster-block/commit/760742763686d561f03768345d672935d72e49b0

i.e. we initially had volserver option to 'gluster-block cli process', we removed it because @vbellur and @pranithk thought it is complicating the cli syntax and schematics.

@amarts My take home from the discussion we had for volserver option was, not many users like complicated cli options.

So, IMO, we should teak the cli something like:

```
Current cli create options:

# gluster-block help
gluster-block (0.3)
usage:
  gluster-block <command> <volname[/blockname]> [<args>] [--json*]

commands:
  create  <volname/blockname> [ha <count>]
                              [auth <enable|disable>]
                              [prealloc <full|no>]
                              [storage <filename>]
                              [ring-buffer <size-in-MB-units>]
                              <host1[,host2,...]> [size]
        create block device [defaults: ha 1, auth disable, prealloc no, size in bytes,
                             ring-buffer default size dependends on kernel]
[...]
```
```
i.e.


From:
# gluster-block  create  <bhv/blockname> ... <bvHOST1[,bvHOST2,...]>

To:
# gluster-block  create  <bhv/blockname> ... [bhvHOST1,bhvHOST2,...@]<bvHOST1[,bvHOST2,...]>
```

Some random examples, to convey the true meaning of above:

Ex1: 
```
when gluster pool is outside the block hosts cluster [ with single BHV HOST IP)
# gluster-block create sample/block ha 3 192.168.1.1@10.70.47.1,10.70.47.2,10.70.47.3 size 1GiB
```
Ex2:
```
when gluster pool is outside the block hosts cluster [ with mutiple BHV HOST IPs)
# gluster-block create sample/block ha 3 192.168.1.1,192.168.1.2@10.70.47.1,10.70.47.2,10.70.47.3 size 1GiB
```

Ex3:
```
when gluster pool is outside the block hosts cluster [ with single BHV HOST IP and other than default port 24007)
# gluster-block create sample/block ha 3 192.168.1.1:24022@10.70.47.1,10.70.47.2,10.70.47.3 size 1GiB
```

Ex4:
```
when gluster pool is within the block hosts cluster
# gluster-block create sample/block ha 3 10.70.47.1,10.70.47.2,10.70.47.3 size 1GiB
```

The delimiters '@', ':' can be chosen differently though.

> 
> > did we thought about, what happens if the supplied host is down ? block won't be able to load right ? SPF ? hence multiple volfile server needs to be supplied.
> 
> Let me give perspective from GCS, where it matters. There will always be a service running, so you will have a hostname, which will resolve and get you a volfile.

Surprisingly, gluster-block doesn't support hostnames. We always depend on fixed IP's.

See:

https://lists.fedorahosted.org/archives/list/targetcli-fb-devel@lists.fedorahosted.org/thread/4E5XQQREDX6S742GRXM7GOLPAEJZQYVX/

https://bugzilla.redhat.com/show_bug.cgi?id=1608478


> 
> > right now there is no way to supply multiple volfile servers to targetcli and parse/format this at tcmu-runner - lets work on this first - I will get back to you on this asap
> > Not an immediate blocker, IMO! As we are looking to deploy it in cloud, there will always be a service, lets solve this later.
> 
> > Also we need some mechanism about getting current set of volfile-servers with glfs api, something like, glfs_get_volfile_server(glfs);
> 
> Because the above is not a requirement, I don't think we should be bothered about this!

If we want to supply one single IP then I see the need for glfs_get_volfile_server(glfs); for sure.

That said, again, can we supply multiple volfile servers and assure one of them is up at any given time ?(think of replace brick case, what if IP's supplied at create time are withdrawn later for maintenance and bricks moved to other nodes)

---

IMO, not much work is needed at targetcli/tcmu-runner to get the multiple volservers accepted and parsed. I will get back with the complete set of changes required at the dependency layers very soon.

Thanks!

amarts:
All below answers are my version of how I envision the changes, happy to discuss further.. Specially I need input from @SaravanaStorageNetwork @raghavendra-talur @humblec @obnoxxx @JohnStrunk 
 can also add their perspective from current day k8s experience.

> gluster-blockd is started by systemd unit today, and how are we going to supply these IP's to the dameon during start time ? Did we thought about this ? even if we want to supply IP's at /etc/sysconfig/gluster-blockd, then top layers will have difficulties automating this.

This patch doesn't change anything with existing eco-system. No IP will be provided to systemd, it will continue to use 'localhost'.

> (BHV -> Block hosting volume)
> Say if we have a BHV1 (IP1, IP2, IP3) and BHV2 (IP4, IP5, IP6) both in different clusters, how you are going to supply the IP's to the dameon now ?

Again, no changes to current behavior, but in 'future' container world, we would have different pod started for dealing with different BHV. That way, we will still have just 1 cluster to deal with, with one daemon.

> @amarts My take home from the discussion we had for volserver option was, not many users like complicated cli options.

I agree, no one likes complex things, but these options would be supplied by installer (say Ansible or operators), and users are not seeing CLI. No changes in CLI is in my plan!

> Surprisingly, gluster-block doesn't support hostnames. We always depend on fixed IP's.

Is that statement true for glfs API? I guess you are talking about iSCSI and block part of it, which **should** remain with IPs. If you see the changes, it is only for glfs_init(), which I believe works fine with hostname.

I am looking at this change to be used only in the new stack of [gcs](https://github.com/gluster/gcs) where glusterfs would be deployed differently than the current way of deploying glusterfs with heketi. Let me know if you have more confusion.

JohnStrunk:
Following up on the hostname/IP discussion (and perhaps I'm misunderstanding)... gd2 has a Service in GCS deployments that provide both a stable DNS and IP to reach GD2. The IP is directed to a currently "Ready" instance of GD2, so no "backup volfile servers" should ever be necessary the primary should always work.

As for the need for a single g-b instance to talk to multiple gluster clusters... It's not needed in GCS, and would, in fact, be undesirable due to upgrades. If you enable block access on a gluster cluster, you should get a dedicated set of g-b targets for that cluster.

rpc/block_svc_routines.c
```diff
@@ -4221,9 +4222,15 @@ block_create_common(blockCreate *blk, char *rbsize, char *prio_path)
...
+  /* If it is known that glusterd is hosted in a
+     given hostname, use that instead */
+  gdHostname = gbConf.glusterdHostname;
+  if (!strcmp(gdHostname, "localhost")) {
+    gdHostname = blk->ipaddr;
+  }
   if (GB_ASPRINTF(&backstore, "%s %s name=%s size=%zu cfgstring=%s@%s%s/%s%s wwn=%s",
                   GB_TGCLI_GLFS_PATH, GB_CREATE, blk->block_name, blk->size,
-                  blk->volume, blk->ipaddr, GB_STOREDIR, blk->gbid,
+                  blk->volume, gdHostname, GB_STOREDIR, blk->gbid,
```
amarts: @pkalever is this enough? Or do we need to do anything further?

pkalever:
@amarts as discussed, lets to stick to the cli create time option.
which would give gluster-block additional flexibility in dealing with multiple gluster pools.

Does this sound good ?

amarts:
> Does this sound good ?

Yes, but I would need some help on that! Not very familiar with the flow in that part.

pkalever:
> > Does this sound good ?
> 
> Yes, but I would need some help on that! Not very familiar with the flow in that part.

Sure @amarts. One of us will take this PR to finish line, and update here.

Thanks!

daemon/gluster-blockd.c
amarts: Done

daemon/gluster-blockd.c
amarts: Done

daemon/gluster-blockd.c
amarts: Done

rpc/glfs-operations.c
amarts: Done

utils/utils.c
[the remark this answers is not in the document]
amarts: Done, now only initialize the variables which are non-zero.

rpc/glfs-operations.c
```diff
@@ -36,7 +38,12 @@ glusterBlockVolumeInit(char *volume, int *errCode, char **errMsg)
     return NULL;
   }
 
-  ret = glfs_set_volfile_server(glfs, "tcp", "localhost", 24007);
+  port = gbConf.glusterdPort;
+  hostname = gbConf.glusterdHostname;
+
+  ret = glfs_set_volfile_server(glfs, "tcp",
+                                (hostname? hostname: "localhost"),
+                                (port? port: 24007));
```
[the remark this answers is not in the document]
amarts: Done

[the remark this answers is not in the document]
amarts: Done

utils/utils.c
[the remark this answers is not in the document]
amarts: Done

amarts:
> right now, kernel doesn't support comma separator for passing multiple IP's

Prasanna, for some time to come, we can live with just 1 hostname/IP to be passed for identifying the glusterd volfile server? That way, can we not live with the existing kernel?

amarts:
> Negative scenarios:

For any scenarios, where new option/value of glusterd IP/hostname is missing, fall back to 'localhost' (or req->ipaddr) like now. If there is any failure, treat it as a genuine failure.

pkalever:
> > right now, kernel doesn't support comma separator for passing multiple IP's
> 
> Prasanna, for some time to come, we can live with just 1 hostname/IP to be passed for identifying the glusterd volfile server? That way, can we not live with the existing kernel?

Yes, This is not a problem for now. Like I said, we also have option of using a different delimiter for now ex: '%' or others.

My concern was about, point 3.

amarts:
> gluster-block cli started to look ugly to me.

IMO, we should not consider 'ugly' as a parameter to measure this, instead consider ease of use, and also notice that this is, in all current possible scenarios, is hidden from user. So, I wouldn't be bothered about this.

pkalever:
> For any scenarios, where new option/value of glusterd IP/hostname is missing, fall back to 'localhost' (or req->ipaddr) like now. If there is any failure, treat it as a genuine failure.

Does this means, you are in agreement at maintaining the /var/lib/gluster-block/bhv-servers.map file for 'volume: volfile-server' mapping ?

lxbsz:
@pkalever  

Will this be compatible with the old command format ?

Or at the same will support both formats ? And if without the volfile server will it fall back to use default IPs ?

If we create using the old format, then will it be okay to delete with new format command ?

vbellur:
> I'm thinking about storing volfileservers at /var/lib/gluster-block/bhv-servers.map as part of create operation, so that for info, delete, modify, list one don't have to mention them explicitly.
> 
> Negative scenarios:
> Thinking about, what if the map file is deleted scenarios?
> We still have to maintain support to provide volfileservers for all other commands (but optional), incase volume is not listed in bhv-servers.map then bailout.

Having a map (or yaml?) file providing hints to gluster-block about servers to contact does seem like a good idea to me. I was also thinking about an input file to reduce CLI clutter.

I also wonder if we can leverage a feature like service discovery exposed by the [Kubernetes services paradigm](https://kubernetes.io/docs/concepts/services-networking/) when gluster-block & glusterfs are deployed as pods in Kubernetes.

pkalever:
> > I'm thinking about storing volfileservers at /var/lib/gluster-block/bhv-servers.map as part of create operation, so that for info, delete, modify, list one don't have to mention them explicitly.
> > Negative scenarios:
> > Thinking about, what if the map file is deleted scenarios?
> > We still have to maintain support to provide volfileservers for all other commands (but optional), incase volume is not listed in bhv-servers.map then bailout.
> 
> Having a map (or yaml?) file providing hints to gluster-block about servers to contact does seem like a good idea to me. I was also thinking about an input file to reduce CLI clutter.

@vbellur there are few issues/challenges with this approach, I have edited my previous comment (https://github.com/gluster/gluster-block/pull/137#issuecomment-438232505) to include them all at one-place, please take a look.

> 
> I also wonder if we can leverage a feature like service discovery exposed by the [Kubernetes services paradigm](https://kubernetes.io/docs/concepts/services-networking/) when gluster-block & glusterfs are deployed as pods in Kubernetes.

The Idea seems like, make sure that our service (deployed in a different namespace) is available via the FQDN from other pods, even if they migrate. Does this also guarantied to work with brick replace as well ?

Also this will be only for kube usecase only.

amarts:
> as now there can be multiple gluster clusters, same name block hosting volumes can also exist. Can we leverage volumeId for better identifying the right volume ?

I thought CSI generates these volumes/block-files with an ID? So, that should provide the uniqueness.

> bhv-servers.map will be updated in local node only while create command, if user runs commands from other nodes, they still need to supply volfileservers

Note that goal in future, with GD2 is there will be just 1 trusted pool, just that there would be more hosts in the cluster. The limitation of GD1 came from the volume scale perspective. So, I believe all we have to solve in version 1 is having a target IP/hostname to reach GD2 service. Lets not plan to be future proof now. Lets solve problem at hand, by solving it for just 1 service IP/hostname. Not even need to handle more than 1 IP in first patch.

> current wait for bricks script won't work (wait-for-bricks.sh), and how do we makesure all the remote bricks are up before starting gluster-blockd ?

Noticed that `wait-for-bricks.sh` uses gluster CLI (gd1), and hence it needs change anyways. We need to fix it, and it should just use a ENV variable as a target, if not set, then localhost. It will be through REST API anyways. Need a separate issue for this in any case.

> there will be side effects on upgrade_activities.sh (mainly for genconfig for now), as it needs to makesure all bricks are up.

Again I see that it uses gluster CLI. need to change regardless of this particular discussion, and its target service should be picked up by a ENV variable.  Need an issue to have gd2 REST API instead here.

> should we be inclining to have a separate metdata volume for maintaining gluster-block metadata ?

I am personally against this idea, as historically we have seen too many issues with keeping data and metadata in different places. Lets not focus on this initially, but will get there if this becomes a need.

> I also wonder if we can leverage a feature like service discovery exposed by the Kubernetes services paradigm when gluster-block & glusterfs are deployed as pods in Kubernetes.

@vbellur we should surely get there by v2 or v3 ?

pkalever:
@amarts 

Okay. we can make progress on this with below assumptions.
1. gluster volume names are unique across the clusters
2. scripts wait-for-bricks.sh and upgrade_activities.sh won't work as expected, and be addressed later as part of different issue/PR.
3. we will persist 'volname: volservers' mapping locally
4. we still have option to supply volserver for all the commands (used if cli is run from other nodes)

pkalever:
@amarts @vbellur @pranithk Have pushed a new PR (https://github.com/gluster/gluster-block/pull/161) which solves Kube and GCS usecase. Please help review. Thanks!

pkalever:
Fixed-by: https://github.com/gluster/gluster-block/pull/161

Issue #140, opened 2018-10-19
warnings seen while building...


amarts:
$ make dist
$ rpmbuild --quiet -ta gluster-block-0.3tar.gz
In file included from common.h:15:0,
                 from common.c:12:
common.c: In function 'glusterBlockParseSize':
common.c:104:11: warning: format '%lld' expects argument of type 'long long int', but argument 3 has type 'ssize_t {aka long int}' [-Wformat=]
       MSG("The size %lld will align to sector size %d bytes\n",
           ^
utils.h:132:29: note: in definition of macro 'MSG'
             fprintf(stdout, fmt, __VA_ARGS__);                       \
                             ^~~
utils.h:178:27: warning: format '%lld' expects argument of type 'long long int', but argument 5 has type 'ssize_t {aka long int}' [-Wformat=]
               fprintf(fd, "[%s] %s: " fmt " [at %s+%d :<%s>]\n",       \
                           ^
common.c:106:7: note: in expansion of macro 'LOG'
       LOG(dom, GB_LOG_ERROR,
       ^~~
common.c:107:38: note: format string is defined here
           "The target device size %lld is will align to the sector size %d",
                                   ~~~^
                                   %ld
utils.c:20:24: warning: missing braces around initializer [-Wmissing-braces]
 struct gbConf gbConf = {LRU_COUNT_DEF, GB_LOG_INFO, GB_LOGDIR, '\0', '\0', '\0', '\0', '\0', '\0', 0};
                        ^
                                                                {                                    }
621 blocks

PR #141, opened 2018-10-19
Fix warnings

amarts:
Fixes: #140


amarts:
Note that this is based on top of #137 as there were same lines edited in these patches. If this gets merged, both the pull requests automatically get closed!

obnoxxx:
I suggest to merge #137 first and then rebase this.

PR #143, opened 2018-10-22
Fix Dockerfile

pkalever:


extras/docker/Dockerfile.buildtest
```diff
@@ -12,7 +12,7 @@ COPY . $BUILDDIR
 # prepare the system
 RUN true \
  && dnf -y install \
-           git autoconf automake gcc libtool make file \
+           git autoconf automake gcc libtool make \
```
obnoxxx: This was actually not a typo. `./configure` prints a warning if `/usr/bin/file` is not found. That said it's not a blocker, since it's only a warning....

extras/docker/Dockerfile.buildtest
```diff
@@ -14,6 +14,7 @@ RUN true \
  && dnf -y install \
            git autoconf automake gcc libtool make \
            glusterfs-api-devel libuuid-devel json-c-devel \
+           tcmu-runner targetcli \
```
obnoxxx: This is not needed for building at all.
Only needed for running gluster-block.
And we can't do runtime tests here because the kernel in travis ubuntu 14.04 is not sufficient, afaik.

pkalever: @obnoxxx probably, you didn't noticed the commit msg/subject-line.

This was an effort to verify if tcmu-runner and targetcli are available, nothing more than that.

obnoxxx: Yeah, I noticed the message. But the build does not require it right now, and the container is never run, so no point in adding it just now. :-)

obnoxxx:
@pkalever Based on the above discussions, if you could update the PR removing the 2 patches to the dockerfile, and keep the moving to `extras/docker`, then I'd be fine with it. Would also like to see it run (and pass) the ci here in github...

pkalever:
@obnoxxx done now. Thanks!

PR #150, opened 2018-11-02
Fix gluster master build

### LINKED ISSUE #75: Use correct gfapi function signatures
This issue came up while installing gluster-block on top of glusterfs 4.1dev. 
The gfapi functions have changed as per the discussion https://github.com/gluster/glusterfs/issues/389# but some gfapi calls inside gluster-block are not updated and are still using the old apis. 
Therefore, update the below functions to use the changed gfapi.

- [ ] glfs_ftruncate

amarts:
Fixes #75 

Also overrides #77 as I had to push one more patch to get the build properly done.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,7 @@ DISTCLEANFILES = Makefile.in gluster-block.spec autom4te.cache
 CLEANFILES = *~ gluster-block.spec
 
 EXTRA_DIST = autogen.sh README.md COPYING-GPLV2 COPYING-LGPLV3 \
-						 gluster-block.spec INSTALL NEWS
+	 gluster-block.spec INSTALL NEWS
 
 if USE_SYSTEMD
 RPM_WITH_FLAGS = --with systemd
--- a/configure.ac
+++ b/configure.ac
@@ -81,7 +81,12 @@ AC_CHECK_HEADERS([stdio.h stdlib.h string.h stdbool.h stddef.h \
 
 # Checks for libraries.
 # glusterfs-api versions have a prefix of "7."
-PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],,
+
+PKG_CHECK_MODULES([GFAPI], [glusterfs-api >= 7.3.6],
+                  [PKG_CHECK_MODULES([gfapi], [glusterfs-api < 7.5.0],
+                                     [AC_DEFINE([GLUSTERFS_NOT_NEW],1,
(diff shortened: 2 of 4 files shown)


pkalever:
@amarts @bhumikagoyal IMO, Guster API (gfapi) shouldn't break the backward compatibility.
This will break all the application, such as qemu, tcmu, samba, gluster-block ... and many more.

IMHO, the fix should go in glusterfs.

Thanks!

lxbsz:
Yeah, agree with Prasanna. And we couldn't assume that only the gluster-block project is using the gfapi.

pkalever:
As this is agreed by everyone in the team, probably we should close it.

@amarts feel free to close it accordingly.

amarts:
> As this is agreed by everyone in the team, probably we should close it.
> 
> @amarts feel free to close it accordingly.

I agree that the 'ideal' fix is glusterfs not changing anything, but as this project depends on the changes done in glusterfs project, safe-guarding the project with a proper version check is not at all a bad option, rather the right way to go.

amarts:
As per the discussions in gluster/glusterfs#539 we need to handle the version number properly in gluster-block project itself. Lets continue reviewing and get this going!

amarts:
@pkalever @lxbsz please take a look. This is important for keeping the project up-to-date with glusterfs-master branch, and there may be more changes. But as it doesn't break anything in existing code, I would prefer to have it. 

If you notice, without this, we are building the project with glusterfs-5.1.x right now in our nightly builds here: https://ci.centos.org/view/Gluster/job/gluster_block_basic/21/console. Ideal is to build it everyday with glusterfs master from previous night. With this patch, we can achieve that.

lxbsz:
> @pkalever @lxbsz please take a look. This is important for keeping the project up-to-date with glusterfs-master branch, and there may be more changes. But as it doesn't break anything in existing code, I would prefer to have it.
> 
> If you notice, without this, we are building the project with glusterfs-5.1.x right now in our nightly builds here: https://ci.centos.org/view/Gluster/job/gluster_block_basic/21/console. Ideal is to build it everyday with glusterfs master from previous night. With this patch, we can achieve that.

Is it possible that in future the API will be updated again, like this ?

If so IMO it will be better that to use a macro that has the version number instead of GLUSTERFS_NOT_NEW, something like GLUSTERFS_VERSION_5_1_X. Will this make sense here ?

Thanks.

amarts:
> something like GLUSTERFS_VERSION_5_1_X. Will this make sense here ?

Agree! We can do something like that.

pkalever:
Fixed by: #178

Issue #151, opened 2018-11-05
list: ignore the vim temporary recovery files


lxbsz:
There is list the hiding files, like:

[root@rhel ~]# gluster-block list dht
block1
block0
block2
block3
.block0.swp
[root@rhel ~]#

PR #152, opened 2018-11-05
list: ignore the vim temporary recovery files

lxbsz:
Ignore the vim temporary recovery files .%.sw*

Fixes: #151

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -2724,11 +2724,7 @@ getSoTgArraysForAllVolume(struct soTgObj *obj, blockGenConfigCli *blk,
     }
 
     while ((entry = glfs_readdir(tgmdfd))) {
-      if (strcmp(entry->d_name, ".") &&
-          strcmp(entry->d_name, "..") &&
-          strcmp(entry->d_name, GB_TXLOCKFILE) &&
-          strcmp(entry->d_name, GB_PRIO_FILENAME)) {
-
+      if (!strchr(entry->d_name, '.')) {
         if (GB_ALLOC(info) < 0) {
           ret = -1;
           goto out;
@@ -4937,10 +4933,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
   }
 
   while ((entry = glfs_readdir (tgmdfd))) {
-    if (strcmp(entry->d_name, ".") &&
-        strcmp(entry->d_name, "..") &&
-        strcmp(entry->d_name, GB_TXLOCKFILE) &&
-        strcmp(entry->d_name, GB_PRIO_FILENAME)) {
+    if (!strchr(entry->d_name, '.')) {
       if (blk->json_resp) {
         json_object_array_add(json_array,
                               GB_JSON_OBJ_TO_STR(entry->d_name));


lxbsz:
Maybe to simplify this we could just ignore all the files started with '.' ?
Cause the creation won't allow the file name started with '.':

```
[root@rhel gluster-block]# gluster-block create dht/.block4 ha 3 prealloc full 192.168.195.162,192.168.195.163,192.168.195.164 1G
block name() should contain only aplhanumeric,'-', '_' characters and should be less than 255 characters long
[root@rhel gluster-block]# 

```

amarts:
> Maybe to simplify this we could just ignore all the files started with '.' ?
> Cause the creation won't allow the file name started with '.':

Yes, looks like that is a good solution IMO. @pkalever what you think?

pkalever:
> Yes, looks like that is a good solution IMO. @pkalever what you think?

TBH, we do not expect users to mount the block hosting volume at the first place. Even if they do for what ever reason, they should not touch /block-meta directory.

But @lxbsz you are right, if someone does it, it is better that we be prepared with some defense logic.

I'm okay to skip the files which start with '.' or even more better would be, skip any file whose name has '.' , that includes prio.info and meta.lock as well :-)

```
diff --git a/rpc/block_svc_routines.c b/rpc/block_svc_routines.c
index f804430..4142799 100644
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -4937,10 +4937,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
   }
 
   while ((entry = glfs_readdir (tgmdfd))) {
-    if (strcmp(entry->d_name, ".") &&
-        strcmp(entry->d_name, "..") &&
-        strcmp(entry->d_name, GB_TXLOCKFILE) &&
-        strcmp(entry->d_name, GB_PRIO_FILENAME)) {
+    if (strchr(entry->d_name, ".")) {
       if (blk->json_resp) {
         json_object_array_add(json_array,
                               GB_JSON_OBJ_TO_STR(entry->d_name));
(END)

```


Thanks!

lxbsz:
> I'm okay to skip the files which start with '.' or even more better would be, skip any file whose name has '.' , that includes prio.info and meta.lock as well :-)
> 
> ```
> diff --git a/rpc/block_svc_routines.c b/rpc/block_svc_routines.c
> index f804430..4142799 100644
> --- a/rpc/block_svc_routines.c
> +++ b/rpc/block_svc_routines.c
> @@ -4937,10 +4937,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
>    }
>  
>    while ((entry = glfs_readdir (tgmdfd))) {
> -    if (strcmp(entry->d_name, ".") &&
> -        strcmp(entry->d_name, "..") &&
> -        strcmp(entry->d_name, GB_TXLOCKFILE) &&
> -        strcmp(entry->d_name, GB_PRIO_FILENAME)) {
> +    if (strchr(entry->d_name, ".")) {
>        if (blk->json_resp) {
>          json_object_array_add(json_array,
>                                GB_JSON_OBJ_TO_STR(entry->d_name));
> (END)
> ```
> Thanks!

Yeah, I agree with this change, it looks better, because the '.' char is not allowed in any place of the target block name, so there won't be any back compatible issue.

```
[root@rhel3 gluster-block]# gluster-block create dht/.block4 ha 3 prealloc full 192.168.195.162,192.168.195.163,192.168.195.164 1G
block name() should contain only aplhanumeric,'-', '_' characters and should be less than 255 characters long
[root@rhel3 gluster-block]# gluster-block create dht/block.4 ha 3 prealloc full 192.168.195.162,192.168.195.163,192.168.195.164 1G
block name() should contain only aplhanumeric,'-', '_' characters and should be less than 255 characters long
[root@rhel3 gluster-block]# 
```

pkalever:
@lxbsz you need to change the commit, as the fix is more generic now :-)

Thanks!

lxbsz:
> @lxbsz you need to change the commit msg, as the fix is more generic now :-)
> 
Yeah, sure, will update it soon. thanks.

lxbsz:
@pkalever Done, please review, thanks.

pkalever:
@lxbsz looks like you need to adopt this change in getSoTgArraysForAllVolume() too.

Thanks!

lxbsz:
> @lxbsz looks like you need to adopt this change in getSoTgArraysForAllVolume() too.
> 
> Thanks!

Yeah, update it again, thanks.

pkalever:
Merged now, Thanks @lxbsz.

PR #157, opened 2018-11-06
gluster-blockd: defend for NULL to avoid crash

lxbsz:


lxbsz:
Updated it.
Thanks.

pkalever:
Merged now. Thanks @lxbsz

PR #159, opened 2018-11-08
replace: update the meta data whenever it fails to create in new node

lxbsz:
When the rpc sending fails, will update the meta data file as CONFIGFAIL, because it has already updated the CONFIGINPROGRESS for the new node.

]# cat block-meta/block2 
VOLUME: dht
GBID: 18130119-da15-49b0-ac59-a15a9e968c1a
HA: 3
ENTRYCREATE: INPROGRESS
PRIOPATH: 192.168.195.162
SIZE: 1073741824
RINGBUFFER: 0
ENTRYCREATE: SUCCESS
192.168.195.164: CONFIGINPROGRESS
192.168.195.162: CONFIGINPROGRESS
192.168.195.163: CONFIGINPROGRESS
192.168.195.162: CONFIGSUCCESS
192.168.195.163: CONFIGSUCCESS
192.168.195.164: CONFIGSUCCESS
192.168.195.162: RPINPROGRESS
192.168.195.160: CONFIGINPROGRESS
192.168.195.163: RPINPROGRESS
192.168.195.164: CLEANUPINPROGRESS
192.168.195.160: CONFIGFAIL                  <<<==== HERE
192.168.195.164: CLEANUPSUCCESS
192.168.195.162: RPSUCCESS
192.168.195.163: RPSUCCESS

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -915,16 +915,15 @@ glusterBlockCreateRemote(void *data)
       LOG("mgmt", GB_LOG_ERROR, "%s hence %s for block %s on "
           "host %s volume %s", strerror(errno), FAILED_REMOTE_CREATE,
           cobj.block_name, args->addr, args->volume);
-      goto out;
     } else if (args->reply) {
       errMsg = args->reply;
       args->reply = NULL;
+      LOG("mgmt", GB_LOG_ERROR, "%s for block %s on host %s volume %s",
+          FAILED_REMOTE_CREATE, cobj.block_name, args->addr, args->volume);
     }
 
     GB_METAUPDATE_OR_GOTO(lock, args->glfs, cobj.block_name, cobj.volume,
                           ret, errMsg, out, "%s: CONFIGFAIL\n", args->addr);
-    LOG("mgmt", GB_LOG_ERROR, "%s for block %s on host %s volume %s",
-        FAILED_REMOTE_CREATE, cobj.block_name, args->addr, args->volume);
 
     ret = saveret;
     goto out;


amarts:
Looks good to me.

lxbsz:
@pkalever 
How about this patches ?

pkalever:
@lxbsz I don't see a scenario where this is useful, can you please help with a testcase or a case where this is needed ?

Thanks!

lxbsz:
> @lxbsz I don't see a scenario where this is useful, can you please help with a testcase or a case where this is needed ?
> 
Without this fixing there won't any problem in the product usage.

The create will fail in two cases:
1, there is something wrong with the RPC itself;
2, the RPC itself works well but there is something wrong with the remote.

Currently we are only updating the metadata with CONFIGFAIL for the case 2.

BTW is this normal and expected ?

Thanks,

pkalever:
CLEANUPINPROGRESS means nothing is touched in remote machine, like remote rpc failed case, since rpc itself is failed, we didn't reach remote procedure on the remote daemon.

CLEANUP_FAIL means we attempt to so something on the remote and failed, so there might be a chance we are half way on the remote, so it definitely needs cleanup.

so for:
> 1, there is something wrong with the RPC itself;

CONFIGINPROGRESS is enough, IMO.

Feel free to correct me.

Thanks.

pkalever:
@lxbsz if you are convinced, please close this. Thanks!

lxbsz:
> CLEANUPINPROGRESS, means nothing is touched on remote machine, like remote rpc failed case, since rpc itself is failed, we didn't reach remote procedure on the remote node.
> 
> CONFIGFAIL, means we attempt to do something on the remote and failed, so there might be a chance we are half way on the remote, so it definitely needs cleanup.

Okay, this makes sense to me.
So I will close it now.
Thanks

> 
> so for:
> 
> > 1, there is something wrong with the RPC itself;
> 
> CONFIGINPROGRESS is enough, IMO.
> 
> Feel free to correct me.
> 
> Thanks.

PR #162, opened 2018-11-14
fix coverity high priority warnings

amarts:
Fixes for coverity 'high' priority warnings as per https://scan.coverity.com/projects/gluster-block_a (Check view defects).

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -603,20 +603,21 @@ glusterBlockCallRPC_1(char *host, void *cobj,
       }
       break;
   }
+  ret = -1;
 
-  ret = reply.exit;
   if (opt != VERSION_SRV) {
     if (GB_STRDUP(*out, reply.out) < 0) {
       goto out;
     }
   } else {
     if (GB_ALLOC(obj) < 0) {
-      return -1;
+      goto out;
     }
     obj->capMax = reply.xdata.xdata_len/sizeof(gbCapObj);
     gbCapObj *caps = (gbCapObj *)reply.xdata.xdata_val;
     if (GB_ALLOC_N(obj->response, obj->capMax) < 0) {
-      return -1;
+      GB_FREE(obj);
+      goto out;
     }
     for (i = 0; i < obj->capMax; i++) {
       GB_STRCPYSTATIC(obj->response[i].cap, caps[i].cap);
@@ -624,6 +625,7 @@ glusterBlockCallRPC_1(char *host, void *cobj,
     }
     *out = (char *) obj;
   }
+  ret = reply.exit;
 
  out:
   if (clnt) {
@@ -4992,7 +4994,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
     GB_ASPRINTF (&errMsg, "Not able to open metadata directory for volume "
                  "%s[%s]", blk->volume, strerror(errCode));
     LOG("mgmt", GB_LOG_ERROR, "glfs_opendir(%s): on volume %s failed[%s]",
-        GB_METADIR, blk->volume, strerror(errno));
+        GB_METADIR, blk->volume, strerror(errCode));
     goto out;
   }
 
--- a/utils/capabilities.c
+++ b/utils/capabilities.c
@@ -45,7 +45,7 @@ gbSetCapabilties(void)
   int ret;
   gbCapObj *caps = NULL;
   char *p, *sep;
-
+  bool free_caps = true;
 
   fp = fopen(GB_CAPS_FILE, "r");
   if (fp == NULL) {
@@ -105,9 +105,10 @@ gbSetCapabilties(void)
   }
 
   globalCapabilities = caps;
+  free_caps = false;
 
  out:
-  if (!globalCapabilities) {
+  if (free_caps) {
     GB_FREE(caps);
   }
   GB_FREE(line);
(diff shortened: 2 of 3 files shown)


utils/utils.h
```diff
@@ -276,12 +276,14 @@ extern struct gbConf gbConf;
...
                 LOG("mgmt", GB_LOG_ERROR,                              \
                     "reading command %s output for %s failed(%s)", tmp,\
                     cmd, strerror(errno));                             \
-              else                                                     \
+                sr->out[0] = '\0';                                     \
+              } else {                                                 \
                 sr->out[newLen++] = '\0';                              \
+              }                                                        \
               sr->exit = blockValidateCommandOutput(sr->out, opt,      \
                                                     (void*)blk);       \
```
lxbsz: Here if fread fails, does it still need to call blockValidateCommandOutput() ?

```
if (ferror( fp ) != 0) {
  [...]
} else {
  sr->out[newLen++] = '\0';
  sr->exit = blockValidateCommandOutput(sr->out, opt,
  [...]
}
```

rpc/block_svc_routines.c
```diff
@@ -611,11 +611,18 @@ glusterBlockCallRPC_1(char *host, void *cobj,
...
     }
     obj->capMax = reply.xdata.xdata_len/sizeof(gbCapObj);
     gbCapObj *caps = (gbCapObj *)reply.xdata.xdata_val;
     if (GB_ALLOC_N(obj->response, obj->capMax) < 0) {
+      GB_FREE(obj);
+      if (res) {
+        freeaddrinfo(res);
+      }
       return -1;
     }
```
pkalever: At both the places we need to call 'goto out;' coz we should also call out clnt_destroy()
Also 'ret = reply.exit;' needs to be moved to later.

utils/capabilities.c
```diff
@@ -104,10 +104,15 @@ gbSetCapabilties(void)
     goto out;
   }
 
+  if (globalCapabilities) {
+    GB_FREE(globalCapabilities);
+  }
+
```
pkalever: I won't object, but IMO this is not needed, as we only user of global var 'globalCapabilities' is in gbSetCapabilties() and this won't be only called once.

But nothing wrong in satisfying coverity requirement :-)

utils/utils.h
pkalever: +1 Xiubo

just set sr->exit = -1; and move on.

utils/capabilities.c
amarts: Done

rpc/block_svc_routines.c
amarts: Done

rpc/block_svc_routines.c
```diff
@@ -626,6 +626,7 @@ glusterBlockCallRPC_1(char *host, void *cobj,
   }
 
  out:
+  ret = reply.exit;
```
pkalever: This should be moved prior to label.
Also, make sure failed exists should reflect 'ret = -1;' as I see some reset of 'ret' set at initialization.

rpc/block_svc_routines.c
```diff
@@ -4990,7 +4991,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
   if (!tgmdfd) {
     errCode = errno;
     GB_ASPRINTF (&errMsg, "Not able to open metadata directory for volume "
-                 "%s[%s]", blk->volume, strerror(errCode));
+                 "%s[%s]", blk->volume, strerror(errno));
```
pkalever: GB_ASPRINTF macro can override errno set by glfs_opendir(). That is why we are using strerror(errCode);

Anything wrong here ? what is the coverity msg you saw ?

utils/capabilities.c
```diff
@@ -105,9 +105,10 @@ gbSetCapabilties(void)
   }
 
   globalCapabilities = caps;
+  free_caps = false;
 
  out:
-  if (!globalCapabilities) {
+  if (free_caps) {
```
pkalever: I didn't get what extra are we solving here ?
previous statement was, if globalCapabilities is NULL, then free caps. globalCapabilities will be NULL in all fail exits of this function.

amarts: Coverity doesn't understand global variable, and the flow that the given function gets called only once. So, it assumes global variable is set, and now the function is called again, and hence it traces all the paths where it can see leaks.

Yes, this is mostly dummy change functionality wise, but solves a coverity report.

rpc/block_svc_routines.c
amarts: ok.

rpc/block_svc_routines.c
amarts: the wrong thing is, errCode is of type int, and it is checked for < 0 few lines below. Which means, for coverity, if errCode is -ve, we shouldn't be passing the value to strerror().

errno getting lost is valid concern, and I guess I will solve only that here.

rpc/block_svc_routines.c
```diff
@@ -5032,7 +5034,7 @@ block_list_cli_1_svc_st(blockListCli *blk, struct svc_req *rqstp)
         GB_METADIR, blk->volume, strerror(errno));
   }
 
-  if (errCode < 0) {
+  if (errCode != 0) {
```
pkalever: @amarts This was intentionally set to '< 0'
Some of the functions might set errCode to -1, if that is the case just reset it to GB_DEFAULT_ERRCODE (because errCode shouldn't be negative), Else carry forward the actual errCode returned, like ENOMEM, EIO, ENOENT etc

pkalever:
@amarts Except that one comment, rest all looks good to me. Thanks!

PR #169, opened 2018-11-29
cli: push output to stderr if remote rpc return non-zero

pkalever:
This is basically what we have missed at #121


pkalever:
@lxbsz @amarts please review.

pkalever:
Updated the Reviewed-by tag, will merge this as soon as the CI passes. Thanks @lxbsz

PR #170, opened 2018-12-04
loadConfig: retry opening conf-file if the initial attempt fail

pkalever:
In-case if your editor (vim) follows write to a new file (.swp, .tmp ..) and
move it to actual file name later. There is a window, where we encounter
'No such file or directory' error on open, because of the editors savefile
transaction.

This patch, simply give few more retries to load the conf file.


pkalever:
@lxbsz PTAL

PR #173, opened 2019-01-23
cli-timeout: make rpc-timeout as configurable option

pkalever:
There are two rpc calls involved in the design,
Channel A: From cli process to local daemon (cli thread)
Channel B: From local daemon (cli thread) to remote daemon (remote thread)

Timeout for channel A is configurable now, using
/etc/sysconfig/gluster-blockd, by adjusting option 'GB_CLI_TIMEOUT' (in seconds)

$ cat /etc/sysconfig/gluster-blockd
GB_CLI_TIMEOUT=900

Changes to this value takes effect every time we start running/triggering new
cli command/request.

At the moment making Timeout configurable for channel B is not important, it is
hardcoded as 300 seconds in the code for now.                                  

Some more details:
-----------------
It would have been an easy fix calling clnt_control() right after clnt_create(),
unfortunately there is a bug in glibc, which is why we had to hack the rpc
generated code instead of directly calling clnt_control(), see my very old
commit [1] which explains why we had to override (in a hacky-way) the default
generated TIMEOUT.

[1] https://github.com/gluster/gluster-block/commit/1dbbd7d457251afd1f005bf262b5c366642535e2

### DIFF
--- a/rpc/rpcl/Makefile.am
+++ b/rpc/rpcl/Makefile.am
@@ -11,13 +11,15 @@ CLEANFILES = *~ $(BUILT_SOURCES)
 
 block.h: block.x
 	rpcgen -hM -o $(top_builddir)/rpc/rpcl/$@ $^
+	$(SED) -i '0,/#endif/s//#endif\nextern struct timeval TIMEOUT;/' \
+	          $(top_builddir)/rpc/rpcl/$@
 
 block_xdr.c: block.x
 	rpcgen -cM -o $(top_builddir)/rpc/rpcl/$@ $^
 
 block_clnt.c: block.x
 	rpcgen -lM -o $(top_builddir)/rpc/rpcl/$@ $^
-	$(SED) -i 's|TIMEOUT = { 25, 0 }|TIMEOUT = { 300, 0 }|' $(top_builddir)/rpc/rpcl/$@
+	$(SED) -i '/TIMEOUT = { 25, 0 }/d' $(top_builddir)/rpc/rpcl/$@
 
 block_svc.c: block.x
 	rpcgen -mM -o $(top_builddir)/rpc/rpcl/$@ $^
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -57,7 +57,7 @@
 
 # define  GB_METASTORE_RESERVE   10485760   /* 10 MiB reserve for block-meta */
 
-# define  GB_DEF_CONFIGPATH      "/etc/sysconfig/gluster-blockd"; /* the default config file */
+# define  GB_DEF_CONFIGPATH      "/etc/sysconfig/gluster-blockd" /* the default config file */
 
 # define  GB_TIME_STRING_BUFLEN  \
           (4 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 6 + 1 +   5)
@@ -132,6 +132,7 @@
 struct gbConf {
   size_t glfsLruCount;
   unsigned int logLevel;
+  size_t cliTimeout;
   char logDir[PATH_MAX];
   char daemonLogFile[PATH_MAX];
   char cliLogFile[PATH_MAX];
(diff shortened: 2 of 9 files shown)


pkalever:
@lxbsz @amarts  Please review, this should fix [BZ#1665543](https://bugzilla.redhat.com/show_bug.cgi?id=1665543)

Thanks!

humblec:
@pkalever Having a cli option to make it configurable for gluster -block cli would be a good choice for customers and other integration pieces like heketi.

humblec:
>one needs to pass the option on every cli command
>
its not everytime, rather whenever customer want to mention( ex: for volumes with big size). Otherwise `default` should be good enough..

pkalever:
@raghavendra-talur IMHO, it would simply complicate the cli usage.

Adjusting the rpc-timeout in config file to a value Less-than-or-Equal-to (<=) service wait timeout, is good enough.

rpc/rpcl/Makefile.am
```diff
@@ -11,13 +11,14 @@ CLEANFILES = *~ $(BUILT_SOURCES)
 
 block.h: block.x
 	rpcgen -hM -o $(top_builddir)/rpc/rpcl/$@ $^
+	$(SED) -i '/#endif/a extern struct timeval TIMEOUT;' $(top_builddir)/rpc/rpcl/$@
```
lxbsz: Is this normal ?

# grep 'extern struct timeval TIMEOUT' ./* -rn
[...]
./rpc/rpcl/block.h:16:extern struct timeval TIMEOUT;
./rpc/rpcl/block.h:229:extern struct timeval TIMEOUT;
./rpc/rpcl/block.h:288:extern struct timeval TIMEOUT;
./rpc/rpcl/block.h:329:extern struct timeval TIMEOUT;
./rpc/rpcl/block.h:334:extern struct timeval TIMEOUT;
./rpc/rpcl/block.h:337:extern struct timeval TIMEOUT;

pkalever:
> First of all thanks for quickly bringing up this PR! :-)
> Now a few comments:
> 
> * It is in general kind of weird to me to have the CLI parse the server configuration file and read a certain variable from it. It would be much more natural to have a command line switch and/or an environment variable that would control the behaviour of the cli.

The configuration file has some common OPTIONS which should apply for both cli and dameon, such as LOG_DIR, LOG_LEVEL, .... I think they can have a unified configuration file, for sharing. (Corrections are welcome)

If all it needs is an Env variable way of adjusting rpc timeout for containers, I can work on it. But I'm bit hesitant towards cli option for rpc-timeout. 


> * Secondly, for the containerized use case, the config file is pretty useless: you'd have to craft a config file with the desired value into the container build. It's not practical or even easily possible to change that value for future invocations...

I thought we bindmount the configfile and persist the changes in config-file in container usecase :-)

> 
> Hence I would request to re-think the approach for this PR.

sure, I will make changes in a way that, gluster-block cli can look for bash exported GB_CLI_TIMEOUT environment variable, and accept it. Let me know, if that works for you guys.

And heketi can run something like,

$ export GB_CLI_TIMEOUT=900; gluster-block create \<volname\>/\<blockname\> ha 3 prealloc full auth enable \<ip-list\> 2TiB --json-pretty

Thanks!

pkalever:
@pkalever fixed the sed issue you pointed.

@raghavendra-talur @obnoxxx @humblec patch2 contains the env variable way of controlling GB_CLI_TIMEOUT.

PTAL.

Thanks!

lxbsz:
It looks good to me and test it.

Tested-by: Xiubo Li <xiubli@redhat.com>

utils/utils.h
```diff
@@ -574,10 +575,13 @@ typedef struct gbConfig {
   bool isDynamic;
   char *GB_LOG_LEVEL;
   ssize_t GB_GLFS_LRU_COUNT;
+  ssize_t GB_CLI_TIMEOUT;  /* seconds */
 } gbConfig;
 
 int glusterBlockSetLogLevel(unsigned int logLevel);
 
+//int glusterBlockSetCliTimeout(size_t timeout);
+
```
amarts: any reason to keep the commented code? Plan for further changes? OK with removing this too.

pkalever: At the definition of this function, I have marked a TODO, it goes like this:

When we actually want to use gbConf (global configuration buffer) in cli process this will be used. I have plans to work on it in the near future, so just preserving it here.

Thanks!

pkalever:
Thanks @amarts @lxbsz for review and test. Updated the flags and merged this PR now.

PR #174, opened 2019-01-23
client: set rpc timeout to 15 minutes instead of 5 minutes

obnoxxx:
For creating a very large block volume, e.g of size 2TB,
with prealloc=full, the current hard-coded client timeout
of 5 minutes is not enough.

Raising this to 15 minutes as a quick fix for situations
where it is necessary to create such a volume.

## TODO

We still need to figure out if 15 minutes is actually a good value. This was what was floating around...

### DIFF
--- a/rpc/rpcl/Makefile.am
+++ b/rpc/rpcl/Makefile.am
@@ -17,7 +17,7 @@ block_xdr.c: block.x
 
 block_clnt.c: block.x
 	rpcgen -lM -o $(top_builddir)/rpc/rpcl/$@ $^
-	$(SED) -i 's|TIMEOUT = { 25, 0 }|TIMEOUT = { 300, 0 }|' $(top_builddir)/rpc/rpcl/$@
+	$(SED) -i 's|TIMEOUT = { 25, 0 }|TIMEOUT = { 900, 0 }|' $(top_builddir)/rpc/rpcl/$@
 
 block_svc.c: block.x
 	rpcgen -mM -o $(top_builddir)/rpc/rpcl/$@ $^


rpc/rpcl/Makefile.am
```diff
@@ -17,7 +17,7 @@ block_xdr.c: block.x
 
 block_clnt.c: block.x
 	rpcgen -lM -o $(top_builddir)/rpc/rpcl/$@ $^
-	$(SED) -i 's|TIMEOUT = { 25, 0 }|TIMEOUT = { 300, 0 }|' $(top_builddir)/rpc/rpcl/$@
+	$(SED) -i 's|TIMEOUT = { 25, 0 }|TIMEOUT = { 900, 0 }|' $(top_builddir)/rpc/rpcl/$@
```
amarts: I am not sure if 900 is enough when the prealloc is for 2TB. What is the per second rate we have? 1GB/s ? Even then, we need 2000 seconds if the Block PV size is 2 TB. While this one line fix is enough as a fix for the problem, we should make it big enough to solve the problem.

pkalever: Also what if the customer attempts to create a 4TB volume ?
This is exactly the same reason why I wanted to make it dynamically configurable.

obnoxxx:
@humblec @phlogistonjohn @raghavendra-talur FYI

rpc/rpcl/Makefile.am
obnoxxx: And, yeah, this is meant more as an example. We need a good number that will help most people. Essentially, we'll need to do some experiments by hand to create, ftruncate and fallocate/zerofill a file of 500GB 1TB, 2TB etc on gluster and see how long it takes.

pkalever:
@obnoxxx now the PR#173 provides a two ways to configure the CLI RPC timeout. 1. config file 2. Env variable.

Hope that is good enough for both standalone and containerized use of gluster-block.

Thanks!

obnoxxx:
@pkalever as I mentioned before, my intention was different (to have a super minimal hard-coded fix to start with), but now that the more flexible option is merged, there's no point for this PR any more. :-)

PR #179, opened 2019-03-15
utils: fix compile warning

lxbsz:
passing the strlen() of the source string as the destination
length is pointless, and gcc-8 now warns about it:

utils.c: In function ‘gbStrcpy’:
utils.c:432:11: warning: ‘strncpy’ specified bound depends on the
length of the source argument [-Wstringop-overflow=]
     ret = strncpy(dest, src, n);
           ^~~~~~~~~~~~~~~~~~~~~
utils.c:427:16: note: length computed here
     size_t n = strlen(src);
                ^~~~~~~~~~~

block_svc_routines.c:184:7: warning: ‘strncat’ specified bound 1
equals source length [-Wstringop-overflow=]
       strncat(out, " ", 1);
       ^~~~~~~~~~~~~~~~~~~~
block_svc_routines.c:183:7: warning: ‘strncat’ specified bound
depends on the length of the source argument [-Wstringop-overflow=]
       strncat(out, element, strlen(element));
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -428,11 +428,26 @@ gbStrcpy(char *dest, const char *src, size_t destbytes,
     if (n > (destbytes - 1))
       return NULL;
 
-    ret = strncpy(dest, src, n);
-    /* strncpy NULL terminates if the last character is \0.  Therefore
-     * force the last byte to be \0
-     */
+    ret = memcpy(dest, src, n);
     dest[n] = '\0';
+
+    return ret;
+}
+
+
+char *
+gbStrcat(char *dest, const char *src, size_t destbytes,
+         const char *filename, const char *funcname, size_t linenr)
+{
+    char *ret;
+    size_t n = strlen(src);
+    size_t m = strlen(dest);
+
+    if (n > (destbytes - 1))
+      return NULL;
+
+    ret = memcpy(dest + m, src, n);
+    dest[m + n] = '\0';
 
     return ret;
 }
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -357,6 +357,10 @@ extern struct gbConf gbConf;
             gbStrcpy((dst), (src), (destbytes),                      \
                      __FILE__, __FUNCTION__, __LINE__)
 
+# define  GB_STRCAT(dst, src, destbytes)                             \
+            gbStrcat((dst), (src), (destbytes),                      \
+                     __FILE__, __FUNCTION__, __LINE__)
+
 # define  GB_STRCPYSTATIC(dst, src)                                  \
             GB_STRCPY((dst), (src), (sizeof(dst)))
 
@@ -621,6 +625,9 @@ int gbStrdup(char **dest, const char *src,
              const char *filename, const char *funcname, size_t linenr);
 
 char* gbStrcpy(char *dest, const char *src, size_t destbytes,
+               const char *filename, const char *funcname, size_t linenr);
+
+char *gbStrcat(char *dest, const char *src, size_t destbytes,
                const char *filename, const char *funcname, size_t linenr);
 
 void gbFree(void *ptrptr);
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -165,6 +165,7 @@ removeDuplicateSubstr(char **line)
   char *temp = *line;
   char *out;
   char *element;
+  size_t len;
 
 
   if (!temp) {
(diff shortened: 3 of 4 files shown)


lxbsz:
@amarts @pkalever 
Please review and thanks.

pkalever:
@lxbsz @amarts 
strcat and strcmp are bad api's ref: https://bugzilla.redhat.com/show_bug.cgi?id=1471078

We need to fix this with better api calls.

lxbsz:
> @lxbsz @amarts
> strcat and strcmp are bad api's ref: https://bugzilla.redhat.com/show_bug.cgi?id=1471078
> 
> We need to fix this with better api calls.

Fixed and updated :-)
Please review @pkalever
Thanks.

utils/utils.h
```diff
@@ -354,7 +354,11 @@ extern struct gbConf gbConf;
                      __FILE__, __FUNCTION__, __LINE__)
 
 # define  GB_STRCPY(dst, src, destbytes)                             \
-            gbStrcpy((dst), (src), (destbytes),                      \
+            gbStrcpy((dst), (src), (destbytes), strlen(src),         \
+                     __FILE__, __FUNCTION__, __LINE__)
+
+# define  GB_STRCAT(dst, src, destbytes)                             \
+            gbStrcat((dst), (src), (destbytes), strlen(src),         \
```
pkalever: > passing the strlen() of the source string as the destination
> length is pointless, and gcc-8 now warns about it:

@lxbsz but we are doing the same pointless thing right :-)

How about:

```
--- a/utils/utils.c                                 
+++ b/utils/utils.c                                 
@@ -477,7 +477,7 @@ gbStrcpy(char *dest, const char *src, size_t destbytes,                              
     if (n > (destbytes - 1))                       
       return NULL;                                 
                                                    
-    ret = strncpy(dest, src, n);                   
+    ret = memcpy(dest, src, n);                    
     /* strncpy NULL terminates if the last character is \0.  Therefore                                  
      * force the last byte to be \0                
      */                                            
@@ -485,3 +485,17 @@ gbStrcpy(char *dest, const char *src, size_t destbytes,                             
                                                    
     return ret;                                    
 }                                                  
+                                                   
+                                                   
+char *                                             
+gbStrcat(char *dest, const char *src, size_t destbytes,                                                 
+         const char *filename, const char *funcname, size_t linenr)                                     
+{                                                  
+    size_t n = strlen(src);                        
+                                                   
+    if (n > (destbytes - 1))                       
+      return NULL;                                 
+                                                   
+    /* strncat will add the NULL terminates */     
+    return memcpy(dest, src, n);                   
+} 
```

lxbsz: Yeah, this looks good to me. Will update it.
Thanks.

lxbsz:
@pkalever 
Updated done. Thanks.

utils/utils.c
```diff
@@ -428,11 +428,20 @@ gbStrcpy(char *dest, const char *src, size_t destbytes,
...
+
+char *
+gbStrcat(char *dest, const char *src, size_t destbytes, size_t srcbytes,
+         const char *filename, const char *funcname, size_t linenr)
+{
+    if (srcbytes > (destbytes - 1))
+      return NULL;
+
+    /* strncat will add the NULL terminates */
+    return strncat(dest, src, srcbytes);
```
pkalever: @lxbsz I still see the use of same old trick here, can we change this to memcpy as well ?

lxbsz: @pkalever Updated done.
Thanks.

utils/utils.h
```diff
@@ -623,6 +627,10 @@ int gbStrdup(char **dest, const char *src,
 char* gbStrcpy(char *dest, const char *src, size_t destbytes,
                const char *filename, const char *funcname, size_t linenr);
 
+char *
+gbStrcat(char *dest, const char *src, size_t destbytes,
```
pkalever: Will it be possible to align this like other declarations in the file ?

Char *  func();

utils/utils.c
```diff
@@ -428,11 +428,27 @@ gbStrcpy(char *dest, const char *src, size_t destbytes,
...
+         const char *filename, const char *funcname, size_t linenr)
+{
+    char *ret;
+    size_t n = strlen(src);
+    size_t m = strlen(dest);
+
+    if (n > (destbytes - 1))
+      return NULL;
+
+    /* strncat will add the NULL terminates */
```
pkalever: should we remove this comment too ?

utils/utils.h
lxbsz: Sure, will fix it.

utils/utils.c
lxbsz: sure.

lxbsz:
@pkalever 
Done.
Thanks.

pkalever:
@lxbsz Thanks!

PR #178, opened 2019-03-14
version: add the gfapi version check

lxbsz:
Signed-off-by: Xiubo Li <xiubli@redhat.com>

PR #181, opened 2019-03-18
fix compile error

itisravi:
Looks like the glfs_ftruncate api changed to include extra arguments.
This caused the following error while compiling:
glfs-operations.c:215:11: error: too few arguments to function ‘glfs_ftruncate’

Fixed it.

Signed-off-by: Ravishankar N <ravishankar@redhat.com>


pkalever:
@itisravi this looks like a dup of https://github.com/gluster/gluster-block/pull/178

PTAL. Thanks!

itisravi:
Indeed.

PR #184, opened 2019-03-20
routines: fix potential double free

lxbsz:

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -4200,6 +4200,7 @@ block_replace_1_svc_st(blockReplace *blk, struct svc_req *rqstp)
 
   if (GB_ALLOC_N(reply->out, 8192) < 0) {
     GB_FREE(reply);
+    reply = NULL;
     goto out;
   }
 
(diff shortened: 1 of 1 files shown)


lxbsz:
@amarts @pkalever 

Please review, thanks.

NOTE: The 'reply' will be freed in GB_RPC_CALL again without this.

rpc/block_svc_routines.c
```diff
@@ -4200,6 +4200,7 @@ block_replace_1_svc_st(blockReplace *blk, struct svc_req *rqstp)
 
   if (GB_ALLOC_N(reply->out, 8192) < 0) {
     GB_FREE(reply);
+    reply = NULL;
```
pkalever: @lxbsz 
```
void                                                                             
gbFree(void *ptrptr)                                                             
{                                                                                
  int save_errno = errno;                                                        
                                                                                 
                                                                                 
  if(*(void**)ptrptr == NULL) {                                                  
   return;                                                                       
  }                                                                              
                                                                                 
  free(*(void**)ptrptr);                                                         
  *(void**)ptrptr = NULL;                                                        
  errno = save_errno;                                                                                                 
}    

```                 

as part of GB_FREE it is supposed to assign NULL to reply, is anything wrong in above function ?

lxbsz: Okay, I missed this part.
Thanks.

PR #186, opened 2019-03-29
dyn-cfg: Avoid putting large char array on the stack

lxbsz:
If there are to many largr char array on the stack, with smaller
stack size limitation like:

There maybe "SEGFAULT, out of stack" issue.


lxbsz:
Since there to many dependency , will close this and push the new commit to #163.

Issue #149, opened 2018-11-02
block create: default prealloc to full, for better performance


pkalever:
As of now prealloc is off by default. Should we rather default prealloc to full by default for better user experience ? 

@pranithk @vbellur @amarts @nixpanic @obnoxxx @lxbsz thoughts ?

Thanks!

lxbsz:
If on it by default, in the future the thin provision device supporting won't be a problem ? And also, the prealloc to full option will cost a longer time for the device's creation, will this be a problem for some cases like there are thousands of pods/contianers creation/deletion per second ?

nixpanic:
As Xiubo mentions, thin-provisioning is something that must be supportable in the future. Defaulting to pre-allocation is now done through Heketi (I think). Not sure how the integration in GD2 looks like, but that should be flexible as well. Good defaults are important, but there need to be ways to request non-default options.

pranithk:
@pkalever +1 from me. Consdering the only usecase for gluster-block is using heketi which creates gluster-block devices with pre-allocation, that seems like a better default than what we have now.

JohnStrunk:
- What does "better performance" mean?
  - Random or sequential?
  - 1st write or overwrite?
  - Does the storage media affect whether this is of benefit (hdd/ssd)?
- What is the effect on volume create?
  - Does it change the rate of volume creation?
  - Does it change the latency of creating a 1 TB volume?
- How would this change affect our ability to do thin provisioning in GCS through thin_p, loop mounted files, and/or VDO?

inevity:
For EC volume which did not support zerofill op , if run create cmd with prealloc full,the command will fail, 
So it is sensible to let use to supply the value?

nixpanic:
On Thu, Apr 18, 2019 at 01:38:10AM -0700, baul wrote:
> For EC volume which did not support zerofill op , if run create cmd with prealloc full,the command will fail, 
> So it is sensible to let use to supply the value?

EC/disperse does still not support zerofill. I expected that
gluster-block does a fallback when it errors out. This should be the
case for tcmu-runner as well... If some operations trigger
glfs_zerofill(), those will all fail on EC/disperse volymes.

I am not sure if fallbacks are in place everywhere, so it may not be
sufficient to turn off prealloc on creation.

pkalever:
[...]

PR #191, opened 2019-03-29
create: turn 'prealloc full' by default

pkalever:
As of now prealloc is off by default.
This patch change the default prealloc to full.

Fixes: #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ commands:
                               [storage <filename>]
                               [ring-buffer <size-in-MB-units>]
                               <host1[,host2,...]> [size]
-        create block device [defaults: ha 1, auth disable, prealloc no, size in bytes,
+        create block device [defaults: ha 1, auth disable, prealloc full, size in bytes,
 	                     ring-buffer default size dependends on kernel]
 
   list    <volname>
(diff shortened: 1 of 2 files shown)


pkalever:
@amarts @lxbsz please take a look. Thanks!

lxbsz:
> @amarts @lxbsz please take a look. Thanks!

@pkalever Since we are changing the cli export to users/customers, how about add this with one version or minor version update ? Then it will be more easy for the hekti or other upper layer tool to adapt the new changes.

pkalever:
> @pkalever Since we are changing the cli exported to users/customers, how about add this with one version or minor version update ? Then it will be more easy for the hekti or other upper layer tool to adapt the new changes.

Heketi cli by default call cli 'prealloc full' options with gluster-block create. Hence I don't see any issues with it.

If needed we can allow users set their default choice with config file option too, let me know if that is what you want.

Thanks!

pkalever:
> LGTM
> 
> Please make sure to call this out in the release notes.

Sure, will add this. Thanks @amarts !

> 
> Reviewed By: Amar Tumballi [amarts@redhat.com](mailto:amarts@redhat.com)

pkalever:
Updated tags, ready for merge. Thanks!

PR #195, opened 2019-04-01
automake: silent building

lxbsz:
Reduce the buidling redudant info.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,6 +15,12 @@ AC_INIT([gluster-block],
 
 AC_SUBST([PACKAGE_RELEASE], 1)
 
+m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES(yes)])
+if make --help 2>&1 | grep -q no-print-directory; then
+    AM_MAKEFLAGS="$AM_MAKEFLAGS --no-print-directory"
+fi
+AC_SUBST(AM_MAKEFLAGS)
+
 AC_ARG_WITH(systemddir,
             AC_HELP_STRING([--with-systemddir=DIR],
                            [systemd service files in DIR @<:@PREFIX/lib/systemd/system@:>@]),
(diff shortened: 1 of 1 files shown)


lxbsz:
@pkalever @amarts 
Please review.
Thanks.

configure.ac
```diff
@@ -15,6 +15,12 @@ AC_INIT([gluster-block],
 
 AC_SUBST([PACKAGE_RELEASE], 1)
 
+m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES(yes)])
+if make --help 2>&1 | grep -q no-print-directory; then
+    AM_MAKEFLAGS="$AM_MAKEFLAGS --no-print-directory"
+fi
+AC_SUBST(AM_MAKEFLAGS)
+
```
pkalever: @lxbsz I like this change, but with this we cannot see with what all compile/linker flags an individual file is compiled/linked.

I'm okay to keep this as default, as this shows less details to users, but It will be very nice to have an option at configure time to enable/disable this, to help dev at debugging.

Thanks!

lxbsz: @pkalever  Make sense and will fix it.

pkalever: @lxbsz 

Can you also check:
$ ./configure --silent
$ make --silent

and see if that is what we are looking at ?

lxbsz: Yeah, it is. Just want to make this disabled as default, make sense ?

pkalever: Yup! please provide an option like --disable-silent or something meaningful. Thanks!

lxbsz:
@pkalever Done.
There already have one option to enable the verbose build output, please see:

./configure --help
[...]
--disable-silent-rules  verbose build output (undo: "make V=0")
[...]

Thanks.

pkalever:
@lxbsz perfect! Looks good to me.

Reviewed-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

pkalever:
@lxbsz could you please update the tags. This is ready for merge. Thanks!

lxbsz:
> @lxbsz could you please update the tags. This is ready for merge. Thanks!

Done Prasanna, thanks.

pkalever:
Thanks @lxbsz !

PR #194, opened 2019-04-01
docs: create CONTRIBUTING.md

pkalever:
Signed-off-by: Prasanna Kumar Kalever <prasanna.kalever@redhat.com>

PR #196, opened 2019-04-01
docs: create CONTRIBUTING.md

pkalever:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+## How to contribute to gluster-block
+
+#### **Did you find a bug?**
+
+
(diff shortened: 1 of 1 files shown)


pkalever:
Restoring the conversation here: Closing #194 

>> @lxbsz @amarts Please review and help with suggestions.
>>
>> you can have a glance about how it looks like, here:
>> https://github.com/pkalever/gluster-block/blob/contributing-pr/CONTRIBUTING.md
>>
>> Thanks!

> @pkalever
> BTW, should we mention that if adding new features, should the test cases be added at the same time ?
> Thanks.

Good thought, let me add it to the doc. Thanks @lxbsz !

Issue #118, opened 2018-09-14
gluster-blockd should have a runtime check for the needed targetcli version


nixpanic:
In #57 a [testing difficulty was hit](https://github.com/gluster/gluster-block/pull/57#issuecomment-413173462) as it seemed newer versions of gluster-block require an updated targetcli. In order to prevent such problems from happening, a clear error should be reported.

When gluster-blockd starts up, it ideally checks if targetcli is running, and has the correct (or newer) version (either through a shell command, or `targetcli --version`). In case the version is not compatible, a message like this should be logged and gluster-blockd should exit with a failure:

    fatal error: targetcli version >= x.y.z is required, got version a.b.c

Issue #76, opened 2018-04-09
Start adding external dependencies (targetcli/rtslib/configshell) version checks


pkalever:
This will save us from breaking compatibility with older versions of dependencies.

PR #206, opened 2019-04-15
daemon: check for dependencies versions at runtime

pkalever:

defend on minimum targetcli, tcmu-runner versions.

### Does this PR fix issues?

Fixes: #76
Fixes: #118

### Notes for the reviewer

I hope, we are good with,

Requesites:
targetcli   >= 2.1.fb46
tcmu-runner >= 1.1.3

### DIFF
--- a/daemon/gluster-blockd.c
+++ b/daemon/gluster-blockd.c
@@ -26,6 +26,7 @@
 # include  "block.h"
 # include  "block_svc.h"
 # include  "capabilities.h"
+# include  "version.h"
 
 # define   GB_TGCLI_GLOBALS     "targetcli set "                               \
                                 "global auto_add_default_portal=false "        \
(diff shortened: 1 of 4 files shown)


pkalever:
@amarts @lxbsz please review. Thanks!

daemon/gluster-blockd.c
```diff
@@ -366,22 +352,114 @@ gbMinKernelVersionCheck(void)
...
+  return;
+
+ out:
+  GB_FREE(out);
+  LOG("mgmt", GB_LOG_ERROR, "%s",
+      "Please install the recommended dependency version and try again.");
+  exit(EXIT_FAILURE);
+}
+
+
```
amarts: While all the above looks fine, it makes me sad that we need to write so much code in C to do it :-/

No other option though.

pkalever: Indeed :-(

pkalever:
Thanks @lxbsz @amarts for the review. Updated the patches with tags. Merging now.

PR #211, opened 2019-04-17
logger: allow without any variable argument

lxbsz:
Allow call the logger helpers without any variable argument

If there is no any variable argument in the logger helpers, it
will hit the following compile errors:

../utils/utils.h:175:34: error: expected expression before ‘,’ token
                       __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);  \
                                  ^
gluster-blockd.c:268:5: note: in expansion of macro ‘LOG’
     LOG ("mgmt", GB_LOG_ERROR, "Test................");
     ^

More detail please see the GNU doc:
https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

### Does this PR fix issues?
No.

### DIFF
--- a/cli/gluster-block.c
+++ b/cli/gluster-block.c
@@ -35,7 +35,7 @@
 # define  GB_ARGCHECK_OR_RETURN(argcount, count, cmd, helpstr)        \
           do {                                                        \
             if (argcount != count) {                                  \
-              MSG(stderr, "Inadequate arguments for %s:\n%s\n", cmd, helpstr); \
+              MSG(stderr, "Inadequate arguments for %s:\n%s", cmd, helpstr); \
               return -1;                                              \
             }                                                         \
           } while(0)
@@ -156,7 +156,7 @@ glusterBlockCliRPC_1(void *cobj, clioperations opt)
 
   conf = glusterBlockCLILoadConfig();
   if (!conf) {
-      LOG("cli", GB_LOG_ERROR, "%s", "glusterBlockCLILoadConfig() failed");
+      LOG("cli", GB_LOG_ERROR, "glusterBlockCLILoadConfig() failed");
       goto out;
   }
 
@@ -254,9 +254,9 @@ glusterBlockCliRPC_1(void *cobj, clioperations opt)
     }
   } else if (errMsg[0]) {
     LOG("cli", GB_LOG_ERROR, "%s", errMsg);
-    MSG(stderr, "%s\n", errMsg);
+    MSG(stderr, "%s", errMsg);
   } else {
-    MSG(stderr, "%s\n", "Did not receive any response from gluster-block daemon."
+    MSG(stderr, "Did not receive any response from gluster-block daemon."
         " Please check log files to find the reason");
     ret = -1;
   }
@@ -286,7 +286,7 @@ glusterBlockCliRPC_1(void *cobj, clioperations opt)
 static void
 glusterBlockHelp(void)
 {
-  MSG(stdout, "%s",
+  MSG(stdout,
       PACKAGE_NAME" ("PACKAGE_VERSION")\n"
       "usage:\n"
       "  gluster-block [timeout <seconds>] <command> <volname[/blockname]> [<args>] [--json*]\n"
@@ -408,9 +408,9 @@ glusterBlockParseVolumeBlock(char *volumeblock, char *volume, char *block,
   /* part before '/' is the volume name */
   sep = strchr(tmp, '/');
   if (!sep) {
-    MSG(stderr, "argument '<volname/blockname>'(%s) is incorrect\n",
+    MSG(stderr, "argument '<volname/blockname>'(%s) is incorrect",
         volumeblock);
-    MSG(stderr, "%s\n", helpstr);
+    MSG(stderr, "%s", helpstr);
     LOG("cli", GB_LOG_ERROR, "%s failed while parsing <volname/blockname>", op);
     goto out;
   }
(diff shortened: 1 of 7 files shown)


lxbsz:
@pkalever @amarts 
Please review, Thanks.

cli/gluster-block.c
```diff
@@ -254,9 +254,9 @@ glusterBlockCliRPC_1(void *cobj, clioperations opt)
     }
   } else if (errMsg[0]) {
     LOG("cli", GB_LOG_ERROR, "%s", errMsg);
-    MSG(stderr, "%s\n", errMsg);
+    MSG(stderr, "%s", errMsg);
```
pkalever: No need of '%s' in MSG() according to your change, I see you just removed '\n' in many more places, please remove %s in all of them.

lxbsz: Sorry, I may not understand this.

BTW, we are also having this:

./utils/common.c:97:        MSG(stderr, "minimum acceptable block size is %d bytes\n", GB_DEFAULT_SECTOR_SIZE);

And the 'errMsg' is the variable argument, not the format string, as far as I know only the following type we could remove the "%s":

MSG(stderr, "%s", "blabla......");  --> MSG(stderr, "blabla....");  // has fmt without any variable argument.

Make sense ?

Thanks.
BRs

pkalever: Sorry I have commented at a wrong line.

cli/gluster-block.c
```diff
@@ -486,8 +486,8 @@ glusterBlockModify(int argcount, char **options, int json)
     if(ret >= 0) {
       mobj.auth_mode = ret;
     } else {
-      MSG(stderr, "%s\n", "'auth' option is incorrect");
-      MSG(stderr, "%s\n", GB_MODIFY_HELP_STR);
+      MSG(stderr, "'auth' option is incorrect");
+      MSG(stderr, "%s", GB_MODIFY_HELP_STR);
```
pkalever: I mean here:

No need of '%s' in MSG() according to your change, I see you just removed '\n' in many more places, please remove %s in all of them.

We can change this to:
MSG(stderr, GB_MODIFY_HELP_STR);

lxbsz: Yeah, make sense.
Will fix it.
Thanks.

lxbsz:
@pkalever 
Updated it. Thanks.

pkalever:
@lxbsz 

Need some more corrections, need to remove '\n' in below places:

```
[root@Fedora-Server gluster-block]# git grep -n MSG | grep "%s\\\n"
cli/gluster-block.c:38:              MSG(stderr, "Inadequate arguments for %s:\n%s\n", cmd, helpstr); \
cli/gluster-block.c:413:    MSG(stderr, "%s\n", helpstr);
cli/gluster-block.c:472:    MSG(stderr, "Inadequate arguments for modify:\n%s\n", GB_MODIFY_HELP_STR);
cli/gluster-block.c:497:      MSG(stderr, "unknown/unsupported option '%s' for modify auth:\n%s\n",
daemon/gluster-blockd.c:268:    MSG(stderr, "%s\n", errMsg);
daemon/gluster-blockd.c:289:      MSG(stderr, "unknown option: %s\n", options[optind-1]);
daemon/gluster-blockd.c:308:      MSG(stdout, "%s\n", argp_program_version);

```

And remove '%s' at daemon/gluster-blockd.c +106 i.e in glusterBlockDHelp()

lxbsz:
@pkalever 
Updated it.
Thanks.

utils/utils.h
```diff
@@ -172,7 +172,7 @@ extern struct gbConf gbConf;
               logTimeNow(timestamp, GB_TIME_STRING_BUFLEN);            \
               fprintf(fd, "[%s] %s: " fmt " [at %s+%d :<%s>]\n",       \
                       timestamp, LogLevelLookup[level],                \
-                      __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);  \
+              ##__VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);        \
```
pkalever: @lxbsz 
Did we disturb this alignment ?
Or you just worried that this
##__VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);\
is not okay ?

Except this, rest all looks good to me.

lxbsz: @pkalever 

Fixed it. Thanks.

pkalever:
@lxbsz Opps!
I remember now, We didn't put '\n' in MSG() and kept it only in LOG() coz, MSG() macros are used for printing output.

With this change: we always get an extra new line in every cli operation output:

```
[root@Fedora-Server gluster-block]# gluster-block create hosting-volume/blockx4 ring-buffer 64 192.168.124.147 1MiB   
IQN: iqn.2016-12.org.gluster-block:6e7bfdab-de94-4eb3-bf1e-64b0566a88aa                                               
PORTAL(S):  192.168.124.147:3260                           
RESULT: SUCCESS              

```

pkalever:
@lxbsz sorry for jumping the gun.
Are you okay to resubmit a new PR by just removing '\n' changes in MSG Macro, if I revert this PR ?

pkalever:
@lxbsz I have force updated the master by reverting the patch, coz it causes regression.
Please attach the patch here or submit a new PR.

Thanks!

lxbsz:
@pkalever 
Sure, do you mind add one MSG_OUT to handle the reply->out only ?

Thanks.

pkalever:
> @pkalever
> Sure, do you mind adding one MSG_OUT() to handle the reply->out only ?
> 
> Currently only the reply->out does not need one extra "\n", all the others need.

@lxbsz agree!

Hmm, interesting!!! 
How about adding some defense code in the MSG() macro ?

```
if the string ends with '\n'
   do not add '\n' to original string
else
   add '\n' to the original string

```

lxbsz:
> > @pkalever
> > Sure, do you mind adding one MSG_OUT() to handle the reply->out only ?
> > Currently only the reply->out does not need one extra "\n", all the others need.
> 
> @lxbsz agree!
> 
> Hmm, interesting!!!
> How about adding some defense code in the MSG() macro ?
> 
> ```
> if the string ends with '\n'
>    do not add '\n' to original string
> else
>    add '\n' to the original string
> ```

Yeah, this looks good to me, will try to this.

Thanks.

PR #213, opened 2019-04-17
logger: allow without any variable argument [new]

lxbsz:
Allow call the logger helpers without any variable argument

If there is no any variable argument in the logger helpers, it
will hit the following compile errors:

../utils/utils.h:175:34: error: expected expression before ‘,’ token
                       __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);  \
                                  ^
gluster-blockd.c:268:5: note: in expansion of macro ‘LOG’
     LOG ("mgmt", GB_LOG_ERROR, "Test................");
     ^

More detail please see the GNU doc:
https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html

### Does this PR fix issues?
No.

### DIFF
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -124,9 +124,19 @@
 
 # define  MSG(fd, fmt, ...)                                          \
           do {                                                       \
+            int len;                                                 \
+            char *buf;                                               \
             if (fd <= 0)       /* including STDIN_FILENO 0 */        \
               fd = stderr;                                           \
-            fprintf(fd, fmt, __VA_ARGS__);                           \
+            len = GB_ASPRINTF(&buf, fmt, ##__VA_ARGS__);             \
+            if (len != -1) {                                         \
+              if (buf[len - 1] != '\n')                              \
+                buf[len - 1] = '\n';                                 \
+              fprintf(fd, "%s", buf);                                \
+              free(buf);                                             \
+            } else {                                                 \
+              fprintf(fd, fmt "\n", ##__VA_ARGS__);                  \
+            }                                                        \
           } while (0)
 
 
@@ -172,7 +182,7 @@ extern struct gbConf gbConf;
               logTimeNow(timestamp, GB_TIME_STRING_BUFLEN);            \
               fprintf(fd, "[%s] %s: " fmt " [at %s+%d :<%s>]\n",       \
                       timestamp, LogLevelLookup[level],                \
-                      __VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);  \
+                      ##__VA_ARGS__, __FILE__, __LINE__, __FUNCTION__);\
               if (fd != stderr)                                        \
                 fclose(fd);                                            \
             }                                                          \
@@ -225,7 +235,7 @@ extern struct gbConf gbConf;
               ret = -1;                                                 \
               goto label;                                               \
             }                                                           \
-            if (GB_ASPRINTF(&write, __VA_ARGS__) < 0) {                 \
+            if (GB_ASPRINTF(&write, ##__VA_ARGS__) < 0) {               \
               ret = -1;                                                 \
             }                                                           \
             if (!ret) {                                                 \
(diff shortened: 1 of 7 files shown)


utils/utils.h
```diff
@@ -124,9 +124,15 @@
 
 # define  MSG(fd, fmt, ...)                                          \
           do {                                                       \
+            int len;                                                 \
+            char buf[8192] = {0};                                    \
```
pkalever: What if there are 10000 block volumes and we run '# gluster-block list <BHV> --json-pretty' :-)
We cannot predict this buffer size @lxbsz .

lxbsz: Yeah, how about malloc a larger one dynamically ?

pkalever: Allocating memory for this per command basis seems unwanted for me, I'm sure you must be thinking the same :-)
Let me think on this tomorrow and get back to you.

Thanks!

lxbsz: Yeah, sure.
This was also why I wanted to add one MSG_OUT instead.

pkalever: 1. How about using GB_ASPRINTF ? (which involves copy)
2. or we may have to use va_start(), va_end() to parse __VA_ARGS__ one by one our self (this may be complicating)
3. or lets not complicate this, lets use MSG_OUT (Why have one more macro for achieving same thing ?)
4. or just follow what was there before (this is generic and lets caller decide what to do)

I will leave it to you :-)

Thanks!

lxbsz: > 1. How about using GB_ASPRINTF ? (which involves copy)
This seems work well and looks good to me.

> 2. or we may have to use va_start(), va_end() to parse **VA_ARGS** one by one our self (this may be complicating)
> 3. or lets not complicate this, lets use MSG_OUT (Why have one more macro for achieving same thing ?)
> 4. or just follow what was there before (this is generic and lets caller decide what to do)
> 
For this IMO we'd better make all the LOG/MSG the same, no need to add the extra '\n' for all of them normally, expect user want to add more than one '\n' at the end.
Or it will be confusing why some have and other do not.

Thanks.

> I will leave it to you :-)
> 
> Thanks!

utils/utils.h
```diff
@@ -124,9 +124,17 @@
...
             if (fd <= 0)       /* including STDIN_FILENO 0 */        \
               fd = stderr;                                           \
-            fprintf(fd, fmt, __VA_ARGS__);                           \
+            len = GB_ASPRINTF(&buf, fmt, ##__VA_ARGS__);             \
+            if (len != -1) {                                         \
+              if (buf[len - 1] != '\n')                              \
+                buf[len - 1] = '\n';                                 \
+              fprintf(fd, "%s", buf);                                \
+              free(buf);                                             \
+            }                                                        \
```
lxbsz: @pkalever 

BTW, should we just assume that it doesn't have the '\n' when the GB_ASPRINTF fails and print the logs by using:

fprintf(fd, fmt "\n", ##__VA_ARGS__); 

Or just do nothing as above ?

Thanks.

pkalever: Yeah, obviously something is better than nothing :-)
Lets print what's there in the 'fmt + __VA_ARGS__', when GB_ASPRINTF fail.

Thanks!

lxbsz: Updated it.
Thanks.

pkalever: Looks good to me.
@lxbsz I also feel like changing line 138 to
fprintf(fd, fmt "\n", ##__VA_ARGS__);

won't hurt. Coz not having a newline looks awkward than having two '\n'.

Reviewed-by: Prasanna Kumar Kalever <prasanna.kalever@redhat.com>

pkalever:
Looks good to me.
@lxbsz I also feel like changing line 138 to
fprintf(fd, fmt "\n", ##VA_ARGS);

won't hurt. Coz not having a newline looks awkward than having two '\n'.

Reviewed-by: Prasanna Kumar Kalever prasanna.kalever@redhat.com

lxbsz:
> Looks good to me.
> @lxbsz I also feel like changing line 138 to
> fprintf(fd, fmt "\n", ##VA_ARGS);
> 
> won't hurt. Coz not having a newline looks awkward than having two '\n'.

@pkalever 
Updated it.
Thanks

PR #217, opened 2019-04-22
gluster-block: fix block delete failure

pkalever:

Problem:
-------

TEST : gluster-block delete hosting-volume/block-volume
FAILED ON:   192.168.124.147
SUCCESSFUL ON: None
RESULT: FAIL
line 98 : NOT OK

Solution:
--------

This is because, the below command was failing,

$ targetcli /backstores/user:glfs delete name=block-volume save=True
Unexpected keyword parameter 'save'.

Since, gluster-block now use per storage object save feature of targetcli,

need, below commit from targetcli, to fix delete failures:
```

commit bca03ea400a691e4218113b6d10adb61ce0cd511 (refs/remotes/origin/save_del)
  Date:   Fri Jun 1 16:15:22 2018 +0530

      saveconfig: way for block-level save with delete comman


```
which is part of targetcli-2.1.fb49, hence turning the dependency check to
latest released and packaged version.

### DIFF
--- a/gluster-block.spec.in
+++ b/gluster-block.spec.in
@@ -32,8 +32,9 @@ BuildRequires:    rpcgen
 BuildRequires:    systemd
 %endif
 
-Requires:         tcmu-runner >= 1.0.4
-Requires:         targetcli >= 2.1.fb43
+Requires:         tcmu-runner >= 1.1.3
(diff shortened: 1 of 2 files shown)


pkalever:
Thanks @lxbsz. Merging now.

PR #218, opened 2019-04-22
gluster-block: fix create failure with ring-buffer option

pkalever:

Problem:
-------

TEST : gluster-block create hosting-volume/block-volume ring-buffer 32 192.168.124.184 1MiB
failed to configure on 192.168.124.184 configure failed
RESULT:FAIL
line 125 : NOT OK

Solution:
--------
We found out that the cfgstring containing commas is parsed correctly by
configshell and passed to the kernel without problems.

we used the following cfgstring:
'dev_config=glfs/hosting-volume@192.168.124.184/block-store/ed9ffb77-63b7-4a91-b2cd-f176eb5be847,max_data_area_mb=32'

But then we see that the ",key1=val1" part has been truncated by the kernel

Read More: https://bugzilla.redhat.com/show_bug.cgi?id=1623925#c24

Later, there was a new way to pass such options via targetcli , with:

```
commit ffafd9680228ec5c39cc721f0db50b7babc2c91d (refs/remotes/origin/control_str, refs/pull/origin/109)
Date:   Fri Apr 6 13:19:54 2018 +0530

    create: add a way to set control string

    this patch enable users to supply control option, a string containing comma
    separated control=value kind tuples, that are passed to the kernel control file.

    Example:
    $ targetcli /backstores/user:glfs create blockX1 1073741824 \
        test@192.168.124.227/block-store/e59309bb-d591-4121-a891-e98ff3416446 \
        control="max_data_area_mb=32,hw_max_sectors=256"

```

This patch, adjust to use control="key1=value1" way of passing control commands

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -4515,7 +4515,7 @@ block_create_v2_1_svc_st(blockCreate2 *blk, struct svc_req *rqstp)
 
 
   if (blk->rb_size) {
-    GB_ASPRINTF(&rbsize, ",%s=%d", GB_RING_BUFFER_STR, blk->rb_size);
+    GB_ASPRINTF(&rbsize, " control='%s=%d'", GB_RING_BUFFER_STR, blk->rb_size);
   }
 
   convertTypeCreate2ToCreate(blk, &blk_v1);


rpc/block_svc_routines.c
```diff
@@ -4515,7 +4515,7 @@ block_create_v2_1_svc_st(blockCreate2 *blk, struct svc_req *rqstp)
 
 
   if (blk->rb_size) {
-    GB_ASPRINTF(&rbsize, ",%s=%d", GB_RING_BUFFER_STR, blk->rb_size);
+    GB_ASPRINTF(&rbsize, " control='%s=%d'", GB_RING_BUFFER_STR, blk->rb_size);
```
lxbsz: BTW, why not just removing the white space before "control=" and then add it in LINE 4037 to be more readble ?

```
4269 blockResponse *
4270 block_create_common(blockCreate *blk, char *rbsize, char *volServer, char *prio_path)
4271 {
....
4307   if (GB_ASPRINTF(&backstore, "%s %s name=%s size=%zu cfgstring=%s@%s%s/%s%s wwn=%s",
4308                   GB_TGCLI_GLFS_PATH, GB_CREATE, blk->block_name, blk->size,
4309                   blk->volume, volServer?volServer:blk->ipaddr, GB_STOREDIR,
4310                   blk->gbid, rbsize ? rbsize: "", blk->gbid) == -1) {
4311     goto out;
4312   }
....
}
```

Will this make sense ?

pkalever: That will add an extra white space every time create command is called without ring-buffer option (which is quite common case).

I just avoid this by pushing the space before ' control='.

Makesense @lxbsz ?

lxbsz: Yeah, that's okay for me :-)

pkalever:
@lxbsz Thanks! Merging now.

PR #223, opened 2019-04-24
zero-fill: use slow zeroing if glfs_zerofill is not supported

pkalever:

Problem:
-------
In case of block hosting volume types like EC/disperse,
glfs_zerofill is not supported, since prealloc is ON by default,
block volume creates will fail.

Solution:
--------
In case if glfs_zerofill() fails with ENOTSUP, perform a
slow write of zeros to the file.

### Does this PR fix issues?

Fixes: https://github.com/gluster/gluster-block/issues/149#issuecomment-484408869

### Notes for the reviewer
Thanks to @inevity for reporting.
Thanks @nixpanic  and @lxbsz for your thoughts about the solution.

### DIFF
--- a/rpc/glfs-operations.c
+++ b/rpc/glfs-operations.c
@@ -14,6 +14,7 @@
 # include "config.h"
 
 # define  GB_LB_ATTR_PREFIX  "user.block"
+# define  GB_ZEROS_BUF_SIZE  4194304  /* 4MiB */
 
 
 struct glfs *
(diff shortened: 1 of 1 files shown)


rpc/glfs-operations.c
```diff
@@ -119,6 +119,66 @@ glusterBlockCheckAvailableSpace(struct glfs *glfs,
...
+  while (size > chunksize * 4) {
+    len = glfs_writev(tgfd, iov, 4, 0);
+    if (len < 0) {
+      LOG("gfapi", GB_LOG_ERROR,
+          "glfs_writev() failed to write zeros: %s", strerror(errno));
+      goto out;
+    }
+    size -= len;
+  }
+
```
lxbsz: @pkalever 
We may also need to do the following check:

```
If (size == 0) {
    ret = 0;
    goto out;
}
```

rpc/glfs-operations.c
```diff
@@ -119,6 +119,66 @@ glusterBlockCheckAvailableSpace(struct glfs *glfs,
...
+  /* Calculate the rest */
+  len = size / chunksize;
+  rest = size % chunksize;
+
+  iov[len].iov_len = rest;
+  if (glfs_writev(tgfd, iov, len, 0) < 0) {
```
lxbsz: The len here should be ：

`if (glfs_writev(tgfd, iov, rest ? len + 1 : len, 0) < 0) {`

Thanks.

rpc/glfs-operations.c
```diff
@@ -119,6 +119,66 @@ glusterBlockCheckAvailableSpace(struct glfs *glfs,
...
+    iov[i].iov_len = chunksize;
+  }
+
+  if (glfs_lseek(tgfd, offset, SEEK_SET) < 0) {
+    LOG("gfapi", GB_LOG_ERROR,
+        "glfs_lseek() failed: %s", strerror(errno));
+    goto out;
+  }
+
+  while (size > chunksize * 4) {
```
pkalever: @lxbsz and here it should be '>=' ?

lxbsz: @pkalever  Yeah, that's right and it should be.

rpc/glfs-operations.c
pkalever: YEah, I missed this, thanks for the catch.

pkalever:
@lxbsz updated the PR, please check this. Thanks!

rpc/glfs-operations.c
```diff
@@ -119,6 +119,71 @@ glusterBlockCheckAvailableSpace(struct glfs *glfs,
 }
 
 
+int
+glusterBlockZeroFill(struct glfs_fd *tgfd, off_t offset, size_t size)
+{
+  struct iovec iov[4];
+  char *zerodata = NULL;
+  size_t chunksize = 4194304;  //4MiB
```
amarts: I see above not changing anywhere, should it be #define instead? Anyways, compiler takes care of optimizing, would be good to comment on why it is a 'variable' (if there is a reason).

pkalever: @amarts make-sense to me, I wrote the code which chunks this initial assignment by 4 buffers, but later I ruled-out that code, and forget to change this.

Will update this in the next spin. Thanks!

pkalever:
@amarts @lxbsz updated now. Thanks!

pkalever:
@lxbsz @amarts Thanks! Merged now.

PR #232, opened 2019-05-05
logrotate: remove postrotate action

lxbsz:
Problem:
---------

Normally we need to capture the SIGHUP in the gluster-blockd daemon,
but since we are fopening and creating the logfiles verytime in LOG(),
so there is no need to do the postrotate action.

Resolution:
------------

Remove the postrotate action in the logrotate config file.


pkalever:
This PR is ready IMO, can you update the tags.

PR #233, opened 2019-05-05
logger: init the logfiles in initGbConfig

lxbsz:

Problem:
---------
Without this, all the logfiles will be initialized to NULL as, and
there will be the following errors with the log messages outputed
to stderr:

May 05 18:33:17 rhel3 gluster-blockd[18005]: Error opening log file: No such file or directory
May 05 18:33:17 rhel3 gluster-blockd[18005]: Logging to stderr.

Resolution:
-----------
Initialize the logfiles in initGbConfig together with the logDir.

### DIFF
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -66,7 +66,6 @@ initGbConfig(void)
 
   gbConf->glfsLruCount = LRU_COUNT_DEF;
   gbConf->logLevel = GB_LOG_INFO;
-  snprintf(gbConf->logDir, PATH_MAX, "%s", GB_LOGDIR_DEF);
   gbConf->cliTimeout = CLI_TIMEOUT_DEF;
 
   return 0;
@@ -441,6 +440,24 @@ glusterLogrotateConfigSet(char *logDir)
 }
 
 
+static bool
+isSamePath(const char *path1, const char *path2)
+{
+    struct stat st1 = {0,};
+    struct stat st2 = {0,};
+
+    if (!path1 || !path2) {
+        return false;
+    }
+
+    if (stat(path1, &st1) == -1 || stat(path2, &st2) == -1) {
+        return false;
+    }
+
+    return st1.st_dev == st2.st_dev && st1.st_ino == st2.st_ino;
+}
+
+
 static int
 initLogDirAndFiles(char *newLogDir)
 {
@@ -469,7 +486,7 @@ initLogDirAndFiles(char *newLogDir)
   if (newLogDir) {
     logDir = newLogDir;
     LOCK(gbConf->lock);
-    if (!strcmp(logDir, gbConf->logDir)) {
+    if (isSamePath(logDir, gbConf->logDir)) {
       UNLOCK(gbConf->lock);
       LOG(dom, GB_LOG_DEBUG,
           "No changes to current logDir: %s, skipping it.",
@@ -477,6 +494,8 @@ initLogDirAndFiles(char *newLogDir)
       goto out;
     }
     UNLOCK(gbConf->lock);
+    LOG(dom, logLevel,
+        "trying to change logDir from %s to %s", gbConf->logDir, logDir);
   } else {
     logDir = getenv("GB_LOGDIR");
 
@@ -490,9 +509,6 @@ initLogDirAndFiles(char *newLogDir)
       logDir = GB_LOGDIR_DEF;
     }
   }
-
-  LOG(dom, logLevel,
-      "trying to change logDir from %s to %s", gbConf->logDir, logDir);
 
   if (strlen(logDir) > PATH_MAX - GB_MAX_LOGFILENAME) {
     LOG(dom, GB_LOG_ERROR, "strlen of logDir Path > PATH_MAX: %s", logDir);
--- a/utils/utils.h
+++ b/utils/utils.h
@@ -161,27 +161,42 @@ extern struct gbConf *gbConf;
 
 # define  LOG(str, level, fmt, ...)                                    \
           do {                                                         \
-            FILE *fd;                                                  \
+            FILE *fd = NULL;                                           \
             char timestamp[GB_TIME_STRING_BUFLEN] = {0};               \
             char *tmp;                                                 \
+            bool logFileExist = false;                                 \
             if (GB_STRDUP(tmp, str) < 0)                               \
               fprintf(stderr, "No memory: %s\n", strerror(errno));     \
             LOCK(gbConf->lock);                                        \
             if (level <= gbConf->logLevel) {                           \
-              if (!strcmp(tmp, "mgmt"))                                \
-                fd = fopen (gbConf->daemonLogFile, "a");               \
-              else if (!strcmp(tmp, "cli"))                            \
-                fd = fopen (gbConf->cliLogFile, "a");                  \
-              else if (!strcmp(tmp, "gfapi"))                          \
-                fd = fopen (gbConf->gfapiLogFile, "a");                \
-              else if (!strcmp(tmp, "cmdlog"))                         \
-                fd = fopen (gbConf->cmdhistoryLogFile, "a");           \
-              else                                                     \
+              if (!strcmp(tmp, "mgmt")) {                              \
+                if (gbConf->daemonLogFile[0]) {                        \
+                    fd = fopen (gbConf->daemonLogFile, "a");           \
+                    logFileExist = true;                               \
+                }                                                      \
+              } else if (!strcmp(tmp, "cli")) {                        \
+                if (gbConf->cliLogFile[0]) {                           \
+                    fd = fopen (gbConf->cliLogFile, "a");              \
+                    logFileExist = true;                               \
+                }                                                      \
+              } else if (!strcmp(tmp, "gfapi")) {                      \
+                if (gbConf->gfapiLogFile[0]) {                         \
+                    fd = fopen (gbConf->gfapiLogFile, "a");            \
+                    logFileExist = true;                               \
+                }                                                      \
+              } else if (!strcmp(tmp, "cmdlog")) {                     \
+                if (gbConf->cmdhistoryLogFile[0]) {                    \
+                    fd = fopen (gbConf->cmdhistoryLogFile, "a");       \
+                    logFileExist = true;                               \
+                }                                                      \
+              } else {                                                 \
                 fd = stderr;                                           \
+              }                                                        \
               if (fd == NULL) {                                        \
-                fprintf(stderr, "Error opening log file: %s\n"         \
-                        "Logging to stderr.\n",                        \
-                        strerror(errno));                              \
+                if (logFileExist)                                      \
+                    fprintf(stderr, "Error opening log file: %s\n"     \
+                            "Logging to stderr.\n",                    \
+                            strerror(errno));                          \
                 fd = stderr;                                           \
               }                                                        \
               logTimeNow(timestamp, GB_TIME_STRING_BUFLEN);            \


utils/utils.c
```diff
@@ -66,8 +66,18 @@ initGbConfig(void)
...
+  snprintf(gbConf->daemonLogFile, PATH_MAX, "%s/gluster-blockd.log",
+           gbConf->logDir);
+  snprintf(gbConf->cliLogFile, PATH_MAX, "%s/gluster-block-cli.log",
+           gbConf->logDir);
+  snprintf(gbConf->gfapiLogFile, PATH_MAX, "%s/gluster-block-gfapi.log",
+           gbConf->logDir);
+  snprintf(gbConf->configShellLogFile, PATH_MAX,
+           "%s/gluster-block-configshell.log", gbConf->logDir);
+  snprintf(gbConf->cmdhistoryLogFile, PATH_MAX, "%s/cmd_history.log",
+           gbConf->logDir);
```
nixpanic: Could you also create these log files (empty) when doing `make install`? That way it is easy to find out that they are not packaged/listed in the `gluster-block.spec`.

In the `gluster-block.spec` those files can be marked as `%ghost` so that they are not included in the archive, but will belong to the RPM once they are created.

lxbsz: @nixpanic 
Yeah, this is very cool. Will add this.
Thanks.

lxbsz:
@nixpanic @amarts @pkalever 
Update it, please review. Thanks.

```
[gluster-block]# rpm -qli gluster-block
Name        : gluster-block
Version     : 0.4
Release     : 1
Architecture: x86_64
Install Date: Tue 14 May 2019 09:45:46 AM CST
Group       : Unspecified
Size        : 641718
License     : GPLv2 or LGPLv3+
Signature   : (none)
Source RPM  : gluster-block-0.4-1.src.rpm
Build Date  : Tue 14 May 2019 09:45:43 AM CST
Build Host  : rhel3
Relocations : (not relocatable)
URL         : https://github.com/gluster/gluster-block
Summary     : Gluster block storage utility
Description :
gluster-block is a CLI utility, which aims at making gluster backed block
storage creation and maintenance as simple as possible.
/etc/logrotate.d/gluster-block
/etc/sysconfig/gluster-blockd
/usr/lib/systemd/system/gluster-block-target.service
/usr/lib/systemd/system/gluster-blockd.service
/usr/libexec/gluster-block
/usr/libexec/gluster-block/upgrade_activities.sh
/usr/libexec/gluster-block/wait-for-bricks.sh
/usr/sbin/gluster-block
/usr/sbin/gluster-blockd
/usr/share/doc/gluster-block-0.4
/usr/share/doc/gluster-block-0.4/README.md
/usr/share/licenses/gluster-block-0.4
/usr/share/licenses/gluster-block-0.4/COPYING-GPLV2
/usr/share/licenses/gluster-block-0.4/COPYING-LGPLV3
/usr/share/man/man8/gluster-block.8.gz
/usr/share/man/man8/gluster-blockd.8.gz
/var/lib/gluster-block
/var/lib/gluster-block/gluster-block-caps.info
/var/log/gluster-block
/var/log/gluster-block/cmd_history.log
/var/log/gluster-block/gluster-block-cli.log
/var/log/gluster-block/gluster-block-configshell.log
/var/log/gluster-block/gluster-block-gfapi.log
/var/log/gluster-block/gluster-blockd.log
[gluster-block]# rpm -qfi /var/log/gluster-block/gluster-blockd.log
Name        : gluster-block
Version     : 0.4
Release     : 1
Architecture: x86_64
Install Date: Tue 14 May 2019 09:45:46 AM CST
Group       : Unspecified
Size        : 641718
License     : GPLv2 or LGPLv3+
Signature   : (none)
Source RPM  : gluster-block-0.4-1.src.rpm
Build Date  : Tue 14 May 2019 09:45:43 AM CST
Build Host  : rhel3
Relocations : (not relocatable)
URL         : https://github.com/gluster/gluster-block
Summary     : Gluster block storage utility
Description :
gluster-block is a CLI utility, which aims at making gluster backed block
storage creation and maintenance as simple as possible.

```

utils/utils.c
lxbsz: @nixpanic 
BTW, will the following runtime files make sense here about this ?

/run/gluster-blockd.lock
/run/gluster-blockd.socket

Thanks.

nixpanic: Files under `/run` are normally created through a `/etc/tmpfiles.d` dropin file. Check `man tmpfiles.d` for details. The files under `/run` should be listed as `%ghost` in the .spec too.

lxbsz:
> Looks good to me. Up to you to decide what to do with the files under `/run`. Either add a 2nd commit to this PR, or put it in its own.

Yeah sure, let's do it in another separate commit about this. Need more test with it.
Thanks.

utils/utils.c
lxbsz: Cool, will go through this and as discussed will raise one new PR to fix it.
Thanks.

utils/utils.c
```diff
@@ -66,8 +66,18 @@ initGbConfig(void)
 
   gbConf->glfsLruCount = LRU_COUNT_DEF;
   gbConf->logLevel = GB_LOG_INFO;
-  snprintf(gbConf->logDir, PATH_MAX, "%s", GB_LOGDIR_DEF);
   gbConf->cliTimeout = CLI_TIMEOUT_DEF;
+  snprintf(gbConf->logDir, PATH_MAX, "%s", GB_LOGDIR_DEF);
+  snprintf(gbConf->daemonLogFile, PATH_MAX, "%s/gluster-blockd.log",
+           gbConf->logDir);
```
lxbsz: Will use the GB_LOGDIR_DEF instead of gbConf->logDir here.

lxbsz:
> @lxbsz Can we simplify this ?
> Just opening the logfiles with O_CREAT, should solve it for us.

@pkalever 
The fopen(, "a"), the "a" here will create it if the logfile is not exist.

The root cause is that when calling the LOG() routine, the gbConf->daemonLogFile is still not initialized, it is '\0', so fopen('\0', "a") will hit this errors.

Thanks.

lxbsz:
> > > @lxbsz Can we simplify this ?
> > > @pkalever
> > > The fopen(, "a"), the "a" here will create it if the logfile is not exist.
> > 
> > 
> > The root cause is that when calling the LOG() routine, the gbConf->daemonLogFile is still not initialized, it is '\0', so fopen('\0', "a") will hit this errors.
> 
> AH! Can you add the RC to the commit msg problem details.
>

Sure, will do it.
 
> If so, I think we should first fix the order of logfile initializing and LOG files usage ?
> 
We cannot always guarantee this, it will make the code not that graceful. IMO initialize the default logfiles when allocating the gbConf is a better chioce. 

> Touching the logfile as part of make_install and spec looks ugly to me. And it will be a maintenance code in the future every time we add new logfile.

I can remove this part.

Thanks.

> 
> Thanks!

pkalever:
> > If so, I think we should first fix the order of logfile initializing and LOG files usage ?
> 
> We cannot always guarantee this, it will make the code not that graceful. IMO initialize the default logfiles when allocating the gbConf is a better chioce.
> 

I didn't understand this part completely. And why its not possible to read the logDir (from various ways, as per precedence) and then use LOG() ?

And say for some reason that you see its really not possible, then, we can probably fix it this way,
```
-              if (!strcmp(tmp, "mgmt"))                                \
-                fd = fopen (gbConf->daemonLogFile, "a");               \
-              else if (!strcmp(tmp, "cli"))                            \
+              if (!strcmp(tmp, "mgmt")) {                              \
+                if (!gbConf->daemonLogFile) {                          \
+                  fd = fopen (gbConf->daemonLogFile, "a");             \
+                } else {                                               \
+                  fd = stdout;                                         \
+                }                                                      \
+            } else if (!strcmp(tmp, "cli"))                            \
```

if we use defaults directly, without looking at configfiles or env variables settings, then we will lose them in cases like containers. Coz the default directories /var/log/gluster-block/ is not persistent in containers.

lxbsz:
> > > If so, I think we should first fix the order of logfile initializing and LOG files usage ?
> > 
> > 
> > We cannot always guarantee this, it will make the code not that graceful. IMO initialize the default logfiles when allocating the gbConf is a better chioce.
> 
> I didn't understand this part completely. And why its not possible to read the logDir (from various ways, as per precedence) and then use LOG() ?
> 
> And say for some reason that you see its really not possible, then, we can probably fix it this way,
> 
> ```
> -              if (!strcmp(tmp, "mgmt"))                                \
> -                fd = fopen (gbConf->daemonLogFile, "a");               \
> -              else if (!strcmp(tmp, "cli"))                            \
> +              if (!strcmp(tmp, "mgmt")) {                              \
> +                if (!gbConf->daemonLogFile) {                          \
> +                  fd = fopen (gbConf->daemonLogFile, "a");             \
> +                } else {                                               \
> +                  fd = stdout;                                         \
> +                }                                                      \
> +            } else if (!strcmp(tmp, "cli"))                            \
> ```
> 
> if we use defaults directly, without looking at configfiles or env variables settings, then we will lose them in cases like containers. Coz the default directories /var/log/gluster-block/ is not persistent in containers.

We will still could see the log like:

[gluster-block]# gluster-blockd --version
[2019-05-28 15:01:58.991488] CRIT: trying to change logDir from /var/log/gluster-block to /var/log/gluster-block [at utils.c+495 :<initLogDirAndFiles>]
gluster-block (0.4)
Repository rev: https://github.com/gluster/gluster-block.git
Copyright (c) 2016 Red Hat, Inc. <https://redhat.com/>
gluster-block comes with ABSOLUTELY NO WARRANTY.
It is licensed to you under your choice of the GNU Lesser
General Public License, version 3 or any later version (LGPLv3
or later), or the GNU General Public License, version 2 (GPLv2),
in all cases as published by the Free Software Foundation.
[gluster-block]# 

Thanks.

pkalever:
> We will still could see the log like:
> 
> [gluster-block]# gluster-blockd --version
> [2019-05-28 15:01:58.991488] CRIT: trying to change logDir from /var/log/gluster-block to /var/log/gluster-block [at utils.c+495 :]
> gluster-block (0.4)
> Repository rev: https://github.com/gluster/gluster-block.git
> Copyright (c) 2016 Red Hat, Inc. https://redhat.com/
> gluster-block comes with ABSOLUTELY NO WARRANTY.
> It is licensed to you under your choice of the GNU Lesser
> General Public License, version 3 or any later version (LGPLv3
> or later), or the GNU General Public License, version 2 (GPLv2),
> in all cases as published by the Free Software Foundation.
> [gluster-block]#

We should not execute logdir change, when logDir and gbConf->logDir are same.

and hence should skip below line too:
```
  LOG(dom, logLevel,                                                                                                                                                                                               
      "trying to change logDir from %s to %s", gbConf->logDir, logDir);    
```

lxbsz:
> > We will still could see the log like:
> > [gluster-block]# gluster-blockd --version
> > [2019-05-28 15:01:58.991488] CRIT: trying to change logDir from /var/log/gluster-block to /var/log/gluster-block [at utils.c+495 :]
> > gluster-block (0.4)
> > Repository rev: https://github.com/gluster/gluster-block.git
> > Copyright (c) 2016 Red Hat, Inc. https://redhat.com/
> > gluster-block comes with ABSOLUTELY NO WARRANTY.
> > It is licensed to you under your choice of the GNU Lesser
> > General Public License, version 3 or any later version (LGPLv3
> > or later), or the GNU General Public License, version 2 (GPLv2),
> > in all cases as published by the Free Software Foundation.
> > [gluster-block]#
> 
> We should not execute logdir change, when logDir and gbConf->logDir are same.
> 
> and hence should skip below line too:
> 
> ```
>   LOG(dom, logLevel,                                                                                                                                                                                               
>       "trying to change logDir from %s to %s", gbConf->logDir, logDir);    
> ```

Yeah, that make sense.

But still have other problems like:

[gluster-block]# gluster-blockd --version
[2019-05-28 15:15:27.978420] INFO: Block Hosting Volfile Server Set to: localhost [at utils.c+348 :<fetchGlfsVolServerFromEnv>]
gluster-block (0.4)
Repository rev: https://github.com/gluster/gluster-block.git
Copyright (c) 2016 Red Hat, Inc. <https://redhat.com/>
gluster-block comes with ABSOLUTELY NO WARRANTY.
It is licensed to you under your choice of the GNU Lesser
General Public License, version 3 or any later version (LGPLv3
or later), or the GNU General Public License, version 2 (GPLv2),
in all cases as published by the Free Software Foundation.

lxbsz:
> > > We will still could see the log like:
> > > [gluster-block]# gluster-blockd --version
> > > [2019-05-28 15:01:58.991488] CRIT: trying to change logDir from /var/log/gluster-block to /var/log/gluster-block [at utils.c+495 :]
> > > gluster-block (0.4)
> > > Repository rev: https://github.com/gluster/gluster-block.git
> > > Copyright (c) 2016 Red Hat, Inc. https://redhat.com/
> > > gluster-block comes with ABSOLUTELY NO WARRANTY.
> > > It is licensed to you under your choice of the GNU Lesser
> > > General Public License, version 3 or any later version (LGPLv3
> > > or later), or the GNU General Public License, version 2 (GPLv2),
> > > in all cases as published by the Free Software Foundation.
> > > [gluster-block]#
> > 
> > 
> > We should not execute logdir change, when logDir and gbConf->logDir are same.
> > and hence should skip below line too:
> > ```
> >   LOG(dom, logLevel,                                                                                                                                                                                               
> >       "trying to change logDir from %s to %s", gbConf->logDir, logDir);    
> > ```
> 
> Yeah, that make sense.
> 
> But still have other problems like:
> 
> [gluster-block]# gluster-blockd --version
> [2019-05-28 15:15:27.978420] INFO: Block Hosting Volfile Server Set to: localhost [at utils.c+348 :]
> gluster-block (0.4)
> Repository rev: https://github.com/gluster/gluster-block.git
> Copyright (c) 2016 Red Hat, Inc. https://redhat.com/
> gluster-block comes with ABSOLUTELY NO WARRANTY.
> It is licensed to you under your choice of the GNU Lesser
> General Public License, version 3 or any later version (LGPLv3
> or later), or the GNU General Public License, version 2 (GPLv2),
> in all cases as published by the Free Software Foundation.

Please ignore this. It is caused by my new changes.

lxbsz:
@pkalever 
Done, please review.
Thanks.

utils/utils.c
```diff
@@ -66,7 +66,6 @@ initGbConfig(void)
 
   gbConf->glfsLruCount = LRU_COUNT_DEF;
   gbConf->logLevel = GB_LOG_INFO;
-  snprintf(gbConf->logDir, PATH_MAX, "%s", GB_LOGDIR_DEF);
```
pkalever: Is there any reason for removing this line ?

lxbsz: In both gluster-blockd and gluster-block they will call initLogging() immediately just after initGbConfig(), which will handle the whole logger stuff there, including the logDir's initialization, so this here in initGbConfig() make no sense and will cause the following issue:

[2019-05-28 15:01:58.991488] CRIT: trying to change logDir from /var/log/gluster-block to /var/log/gluster-block [at utils.c+495 :]

pkalever: I doubt I got you right here.

As we are initializing all the gbConf stuff, lets also keep LOGDIR defaults initialization here.

Further,
main()->initLogging()->initLogDirAndFiles(NULL)

```
static int                                                                       
initLogDirAndFiles(char *newLogDir)                                              
{                                                                                
...
  if (newLogDir) {                                                               
...
    LOG(dom, logLevel,                                                           
        "trying to change logDir from %s to %s", gbConf->logDir, logDir);        
  } else {                                                                 
...
```
If this the path, you are talking about, as newLogDir is NULL, this won't hit now.

in case, If you are taking about other route:
glusterBlockSetLogDir->initLogDirAndFiles(logDir)

We should change comparing paths with strcmp(logDir, gbConf->logDir) and adopt:
```
static bool is_same_path(const char* path1, const char* path2)                                                                                                                                                     
{                                                                                
  struct stat st1 = {0,};                                                        
  struct stat st2 = {0,};                                                        
                                                                                 
  if (!path1 || !path2)                                                          
    return false;                                                                
                                                                                 
  if (stat(path1, &st1) == -1 || stat(path2, &st2) == -1) {                      
    return false;                                                                
  }                                                                              
                                                                                 
  return st1.st_dev == st2.st_dev && st1.st_ino == st2.st_ino;                   
}  
``

lxbsz: I am okay with this if you don't like this change :-)
I just want to collect all the logger init stuff into initLogging() routine because it will fallback to the GB_LOGDIR_DEF always if neither of the env and config option is set.

pkalever: @lxbsz I'm okay, moving this to initLogging(), as I see, other logfile initializing is happening there.

But please fix, if both the dirs are same it won't/shouldn't change again. And using strcmp for dirs is not the right thing, as users can give it dir/ or dir or dir//, all pointing to same path.

lxbsz: Yeah, sure.
Fixing it now.
Thanks.

lxbsz:
@pkalever 
Done.
Thanks.

PR #235, opened 2019-05-09
daemon: remove the tcmu-runner active check in systemd case

lxbsz:

Systemd service will help to guarantee that the tcmu-runner is already up when reaching here.

To remove this in system case just want to avoid the gap in tcmu-runner and gluster-blockd services that
when the tcmu-runner service is dead, the gluster-blockd service won't exit in time and will be still active for a while, which will still to check the tcmu-runner daemon and exit by itself.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -28,6 +28,9 @@ AC_ARG_WITH(systemddir,
             [systemddir='${prefix}/lib/systemd/system'])
 AC_SUBST(systemddir)
 AM_CONDITIONAL([USE_SYSTEMD], test [ -d '/usr/lib/systemd/system' ])
+if test -d '/usr/lib/systemd/system'; then
+    AC_DEFINE([USE_SYSTEMD], [1], [use systemd])
+fi
 
 AC_ARG_WITH(initddir,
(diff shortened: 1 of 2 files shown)


daemon/Makefile.am
```diff
@@ -4,6 +4,7 @@ gluster_blockd_SOURCES = gluster-blockd.c
 
 gluster_blockd_CFLAGS = $(GFAPI_CFLAGS) $(TIRPC_CFLAGS)                        \
                         -DDATADIR=\"$(localstatedir)\"                         \
+                        -DENABLE_SYSTEMD=\"$(USE_SYSTEMD)\"                    \
```
nixpanic: This will work, but would it not be nicer to have #defines for features in `config.h`?

lxbsz: @nixpanic 

Yeah, true. And the ENABLE_SYSTEMD maybe used in other c files in future and will fix it.

Thanks.

lxbsz:
@pkalever @nixpanic 
Updated it, please review again.
Thanks.

PR #240, opened 2019-06-13
alua: disable the usless ALUA state

lxbsz:

Currently we are only using the ALUA's Primary Target Port Group
with implicit transition, which will only use the active/optimized
and active/non-optimized states, so all the others should be disabled
here. Or it will report the wrong state to initiator like:

Report target port groups:
  target port group id : 0x2 , Pref=0
    target port group asymmetric access state : 0x01
    T_SUP : 1, O_SUP : 1, LBD_SUP : 0, U_SUP : 1, S_SUP : 1, AN_SUP : 1, AO_SUP : 1
    ...
  target port group id : 0x1 , Pref=0
    target port group asymmetric access state : 0x00
    T_SUP : 1, O_SUP : 1, LBD_SUP : 0, U_SUP : 1, S_SUP : 1, AN_SUP : 1, AO_SUP : 1
    ...

### DIFF
--- a/rpc/block_svc_routines.c
+++ b/rpc/block_svc_routines.c
@@ -2685,11 +2685,17 @@ getSoObj(char *block, MetaInfo *info, blockGenConfigCli *blk)
 
     json_object_object_add(so_obj_alua_ao_tpg, "alua_access_type", json_object_new_int(1));
     json_object_object_add(so_obj_alua_ao_tpg, "alua_access_state", json_object_new_int(0));
+    json_object_object_add(so_obj_alua_ao_tpg, "alua_support_offline", json_object_new_int(0));
+    json_object_object_add(so_obj_alua_ao_tpg, "alua_support_standby", json_object_new_int(0));
+    json_object_object_add(so_obj_alua_ao_tpg, "alua_support_unavailable", json_object_new_int(0));
     json_object_object_add(so_obj_alua_ao_tpg, "name", GB_JSON_OBJ_TO_STR(GB_ALUA_AO_TPG_NAME));
     json_object_object_add(so_obj_alua_ao_tpg, "tg_pt_gp_id", json_object_new_int(1));
 
     json_object_object_add(so_obj_alua_ano_tpg, "alua_access_type", json_object_new_int(1));
     json_object_object_add(so_obj_alua_ano_tpg, "alua_access_state", json_object_new_int(1));
+    json_object_object_add(so_obj_alua_ano_tpg, "alua_support_offline", json_object_new_int(0));
+    json_object_object_add(so_obj_alua_ano_tpg, "alua_support_standby", json_object_new_int(0));
+    json_object_object_add(so_obj_alua_ano_tpg, "alua_support_unavailable", json_object_new_int(0));
     json_object_object_add(so_obj_alua_ano_tpg, "name", GB_JSON_OBJ_TO_STR(GB_ALUA_ANO_TPG_NAME));
     json_object_object_add(so_obj_alua_ano_tpg, "tg_pt_gp_id", json_object_new_int(2));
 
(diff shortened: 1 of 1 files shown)


lxbsz:
@pkalever 
Please review this one.
Thanks.

pkalever:
@lxbsz This makesense to me. Let me test this and get back.

Thanks!

lxbsz:
@pkalever 
Updated it with small fix.
Thanks.

pkalever:
> @pkalever
> Updated it with small fix.

@lxbsz 
OK. Will retest it now. Thanks!

rpc/block_svc_routines.c
```diff
@@ -4343,6 +4349,45 @@ block_create_common(blockCreate *blk, char *rbsize, char *volServer, char *prio_
...
+
+    /*
+     * The ALUA states and transition:
+     *
+     * a) active/optimized: supported
+     * b) active/non-optimized: supported
+     * c) standby: non-supported
+     * d) unavailable: non-supported
+     * e) offline: non-supported
+     *
```
pkalever: extra line ?

rpc/block_svc_routines.c
```diff
@@ -4343,6 +4349,45 @@ block_create_common(blockCreate *blk, char *rbsize, char *volServer, char *prio_
...
+                    "%s/%s/alua/glfs_tg_pt_gp_ao set alua alua_support_standby=0\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ao set alua alua_support_unavailable=0\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ao set alua alua_support_transitioning=1"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_active_optimized=1\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_active_nonoptimized=1\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_offline=0\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_standby=0\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_unavailable=0\n"
+                    "%s/%s/alua/glfs_tg_pt_gp_ano set alua alua_support_transitioning=1",
+                    GB_TGCLI_GLFS_PATH, blk->block_name,
```
pkalever: I think by default everything is enabled.
So we don't have to explicitly enable alua_support_active_optimized, alua_support_active_nonoptimized and alua_support_transitioning states, this reduces 6 less settings and hence the time to create per block volume will be less impacted.

lxbsz: Not sure, this should be depended with different kernel version.

And there is no need to build the cache for each of them. So for the time shouldn't be much impacted IMO.

Thanks.

rpc/block_svc_routines.c
lxbsz: Will delete it.
Thanks.

rpc/block_svc_routines.c
pkalever: @lxbsz 

Can we have a check on this ? IMHO if weare not changing anything currently about this in the kernel, and atleast since quite sometime now, we should be okay to avoid extra settings.

Also you need to apply the same to genconfig logic as well.

lxbsz: Okay.
Will check and remove it.
Thanks.

lxbsz:
@pkalever 
Updated it and fixed both the genconfig and default value setting issues.
Thanks.

lxbsz:
@pkalever 
Updated the tag.
Thanks.

pkalever:
@lxbsz Merged now. Thanks!

Issue #241, opened 2019-06-19
use 24 hours format UTC time in PrintLogs of wait-for-bricks and upgrade script


pkalever:
### Kind of issue

Bug:

Right now gluster-block-bricks-start.log and gluster-block-upgrade-activities.log prints logs in 12 hours format.

Thanks to @pranithk for pointing this out.

### Observed behavior
PrintLog is using 0..12 i.e. %I instead of 0..23 i.e. %H for printing hours.

### Expected/desired behavior
Logs should be in 24 hours and in UTC

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1720783

PR #242, opened 2019-06-19
Print log

pkalever:

Problem:
---
Right now gluster-block-bricks-start.log and gluster-block-upgrade-activities.log gets log timestamp in 12 hours format.
    
Solution:
---
PrintLog is using 0..11 i.e. %I instead of 0..23 i.e. we should use %H for printing hours.
    

    

### Does this PR fix issues?

Fixes: #241
reported: BZ#1720783

### Notes for the reviewer

Credits to Pranith Kumar Karampuri <pkarampu@redhat.com> for chasing this out.


pkalever:
@pranithk @lxbsz fixed this now, could you please take a look. Thanks!

pkalever:
Updated tags, thanks @pranithk @lxbsz @amarts for the review.

PR #244, opened 2019-07-02
configure.ac: set the libexecdir flexiblly depending on the prefix

lxbsz:

When doing set the --prefix=XXX the GLUSTER_BLOCKD_LIBEXECDIR will be:

GLUSTER_BLOCKD_LIBEXECDIR=NONE/libexec/gluster-block and when installing
it will be installed in extra/NONE/libexec/gluster-block/.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -142,6 +142,8 @@ AC_SUBST(GLUSTER_BLOCKD_WORKDIR)
 if test "x$prefix" = xNONE; then
    test $libexecdir = '${prefix}/libexec' && localstatedir=$ac_default_prefix/libexec
    libexecdir=/usr/libexec
+else
+   libexecdir='${prefix}/libexec'
 fi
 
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -141,7 +141,9 @@ AC_SUBST(GLUSTER_BLOCKD_WORKDIR)
 
 if test "x$prefix" = xNONE; then
    test $libexecdir = '${prefix}/libexec' && localstatedir=$ac_default_prefix/libexec
-   libexecdir=/usr/libexec
+   libexecdir=usr/libexec
```
pkalever: @lxbsz did you intentionally removed '/' before usr/libexec ?

lxbsz: Ah, sorry. This is by mistake.
Fixed it.
Thanks.

lxbsz:
Updated the tag.
Thanks.

pkalever:
Merged now. Thanks @lxbsz

PR #246, opened 2019-07-10
upgrade_activities: always wait till the background daemon is terminated

lxbsz:

When the upgrade activities script finishes and exit successfully,
the gluster-blockd daemon running in the background may still not
killed, so this will be a problem when the gluster-block service
try to start:

ERROR: gluster-blockd is already running...

And also when running the gluster-blockd daemon in background it
may need a while to get fired, so here will always try 5 times by
waiting 1 second each time.

### DIFF
--- a/extras/upgrade_activities.sh
+++ b/extras/upgrade_activities.sh
@@ -19,17 +19,27 @@ GB_TMP_SAVEFILE="/tmp/gb_saveconfig.json"
 USED_COMMANDS="ps sed grep awk head echo pkill gluster gluster-block gluster-blockd"  # append new use of commands here
 mkdir -p "${LOGDIR}"
 GBD_STARTED=0
+GBD_PID=0
+
+
+function kill_gluster_blockd_wait()
+{
+  if [[ ${GBD_STARTED} -ne 0 && ${GBD_PID} -ne 0 ]]
+  then
+    # kill gluster-blockd
+    kill -15 ${GBD_PID} > /dev/null 2>&1
+
+    # let's wait until the child process is finished
+    wait ${GBD_PID}
+  fi
+}
 
 
 function my_exit()
 {
   local ret=${1}
 
-  # kill gluster-blockd
-  if [[ ${GBD_STARTED} -ne 0 ]]
-  then
-    pkill -15 gluster-blockd
-  fi
+  kill_gluster_blockd_wait
 
   exit ${ret}
 }
@@ -81,7 +91,19 @@ function check_daemons_running()
     my_exit 1
   fi
 
-  if ! ps cax | grep -wq '[g]luster-blockd'
+  # try 5 times to make sure the backgroud process is fired.
+  count=1
+  while [[ ${count} -le 5 ]]
+  do
+    if ps cax | grep -wq '[g]luster-blockd'
+    then
+      break
+    fi
+    sleep 1
+    let "count++"
+  done
+
+  if [[ ${count} -gt 5 ]]
   then
     printLog "ERROR: Gluster-blockd is not running"
     my_exit 1
@@ -122,6 +144,7 @@ printLog "INFO: started upgrade activities"
 check_dependent_commands_exist
 # start gluster-blockd
 gluster-blockd --no-remote-rpc & > /dev/null 2>&1
+GBD_PID=${!}
 GBD_STARTED=1
 check_daemons_running
 
@@ -143,6 +166,8 @@ then
     printLog "INFO: successfully completed the upgrade activities"
     my_exit 0
   fi
+  printLog "WARNING: not completely successful with upgrade activities, on next start of gluster-blockd service will try again!"
+  my_exit 1
 fi
 
 # generate the block volumes target configuration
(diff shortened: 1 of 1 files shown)


extras/upgrade_activities.sh
```diff
@@ -29,6 +29,9 @@ function my_exit()
   if [[ ${GBD_STARTED} -ne 0 ]]
   then
     pkill -15 gluster-blockd
+
+    # let's wait until the child process is finished
```
pkalever: There is one white space right after this comment, in both the places.

extras/upgrade_activities.sh
```diff
@@ -29,6 +29,9 @@ function my_exit()
   if [[ ${GBD_STARTED} -ne 0 ]]
   then
     pkill -15 gluster-blockd
+
+    # let's wait until the child process is finished 
+    wait
```
pkalever: Xiubo, should we rather be more specific ?
wait \<pid-of-gluster-blockd\> ?

extras/upgrade_activities.sh
```diff
@@ -81,7 +84,19 @@ function check_daemons_running()
     my_exit 1
   fi
 
-  if ! ps cax | grep -wq '[g]luster-blockd'
+  # try 5 times to make sure the backgroud process is fired.
+  count=1
+  while (($count <= 5))
```
pkalever: In all the places where we use a variable please use ${my_var} syntax, flower braces to maintain consistency across the file.

pkalever: Also, it will be nice to use the same syntax in other places where we check for something like
while [[ ${count} -le 5 ]]
...

if [[ ${count} -gt 5]]
...

extras/upgrade_activities.sh
```diff
@@ -160,6 +175,9 @@ fi
 # kill gluster-blockd
 pkill -15 gluster-blockd
 
+# let's wait until the child process is finished
```
pkalever: white space at end

extras/upgrade_activities.sh
pkalever: This \<pid-of-gluster-blockd\> need to be captured after:

```
# start gluster-blockd                                                                                                                                                                                             
gluster-blockd --no-remote-rpc & > /dev/null 2>&1
GB_PID=${!}
```

Else we might endup getting pidof actual gluster-blockd started by systemd, may be ?

extras/upgrade_activities.sh
lxbsz: Will fix it.
Thanks.

extras/upgrade_activities.sh
lxbsz: This is okay.

The "wait" here will only wait the child processes running in background of this script, here is only the "gluster-blockd --no-remote-rpc &".

But I am okay with adding the PID.

extras/upgrade_activities.sh
lxbsz: Sure, will fix it.

extras/upgrade_activities.sh
lxbsz: Will fix it.

lxbsz:
@pkalever 
Updated them all.
Thanks.

lxbsz:
> @lxbsz BTW - did you check the complete script and tested this with no BHV's and with BHV & BV's in it ?

Yeah, test it with 3 BHVs and some BVs in them.

And test it by deleting all the BHVs again.

All works as expected in my side.

Thanks!

extras/upgrade_activities.sh
```diff
@@ -28,7 +29,13 @@ function my_exit()
   # kill gluster-blockd
   if [[ ${GBD_STARTED} -ne 0 ]]
   then
-    pkill -15 gluster-blockd
+    kill -15 ${GBD_PID} > /dev/null 2>&1
```
pkalever: What's the point in killing if GBD_PID=0 ?

lets check for ${GBD_STARTED} -ne 0 && ${GBD_PID} -ne 0 ? what do you think ?

extras/upgrade_activities.sh
```diff
@@ -158,7 +178,10 @@ else
 fi
 
 # kill gluster-blockd
-pkill -15 gluster-blockd
+kill -15 ${GBD_PID} > /dev/null 2>&1
+
+# let's wait until the child process is finished
+wait ${GBD_PID}
```
pkalever: should we also do the same here ? if GBD_PID -gt 0; kill and wait ?
Or may be create a function like
```
function kill_N_wait()
{
  local pid=${1}

  kill -15 ${pid}
  wait ${pid}
}
```

extras/upgrade_activities.sh
lxbsz: In theory the GBD_PID shouldn't be 0.
But agree to do the check here.

extras/upgrade_activities.sh
lxbsz: I think this is okay, since it is here we can be sure that the daemon is already running.
But it maybe always better to have this check.

lxbsz:
@pkalever 
Fixed them both.
Thanks.

lxbsz:
@pkalever 
Done.
Thanks.

pkalever:
Thanks @lxbsz , merged now.

PR #250, opened 2019-07-26
fix covery issues reported on 0.5dev (master)

pkalever:

Coverity issues reported at:
https://scan.coverity.com/projects/gluster-gluster-block?tab=overview
https://scan4.coverity.com/reports.htm#v43089/p15784

### Notes for the reviewer

* Null pointer dereferences
* thread dead lock
* Error handling

### DIFF
--- a/rpc/block_create.c
+++ b/rpc/block_create.c
@@ -415,7 +415,9 @@ glusterBlockCreateRemoteAsync(blockServerDefPtr list, size_t mpath,
 
  out:
   for (i = 0; i < mpath; i++) {
-    GB_FREE(args[i].reply);
+    if (args) {
+      GB_FREE(args[i].reply);
+    }
   }
   GB_FREE(args);
   GB_FREE(tid);
--- a/utils/utils.c
+++ b/utils/utils.c
@@ -698,7 +698,9 @@ fail:
   GB_FREE(buf);
 
  out:
-  pclose(fp);
+  if (fp) {
+    pclose(fp);
+  }
   return buf;
 }
 
--- a/rpc/glfs-operations.c
+++ b/rpc/glfs-operations.c
@@ -125,7 +125,7 @@ glusterBlockZeroFill(struct glfs_fd *tgfd, off_t offset, size_t size)
 {
   struct iovec iov[4];
   char *zerodata = NULL;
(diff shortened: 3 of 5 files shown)


rpc/block_create.c
```diff
@@ -415,7 +415,9 @@ glusterBlockCreateRemoteAsync(blockServerDefPtr list, size_t mpath,
 
  out:
   for (i = 0; i < mpath; i++) {
-    GB_FREE(args[i].reply);
+    if (args[i]) {
```
lxbsz: For this, it should be:

if (args && args[i]) {

Or just before the 'for (i = 0; i < mpath; i++) {' to check the args ?

utils/utils.c
```diff
@@ -698,7 +698,9 @@ gbRunnerGetOutput(char *cmd)
   GB_FREE(buf);
 
  out:
-  pclose(fp);
+  if (fp) {
+    pclose(fp);
+  }
```
lxbsz: IMO, this is make no much sense, but no harm. It is up to you.

Because the fp = fopen(), will set the fp to NULL or the actual handler.

rpc/block_create.c
```diff
@@ -415,7 +415,9 @@ glusterBlockCreateRemoteAsync(blockServerDefPtr list, size_t mpath,
 
  out:
   for (i = 0; i < mpath; i++) {
-    GB_FREE(args[i].reply);
+    if (args + i) {
```
lxbsz: Will it be a problem if the args == NULL ?
Will if crash ?

rpc/block_create.c
lxbsz: Please ignore this, it is the stale comment.

utils/utils.c
pkalever: Coverity reports it :-)

rpc/block_create.c
pkalever: Grr! makesense, let me fix it.

pkalever: Fixed now, please take look @lxbsz

utils/utils.c
lxbsz: Okay.

PR #253, opened 2019-09-17
ringbuffer: fix the max limited size to 1024M

lxbsz:

The ringbuffer size for each target could up to the global limitation
which is 2GB, here we will allow it up to 1GB. This relevant kernel
patch has been updated into the same verions with the dynamic growth
and shrink patch set.

### Does this PR fix issues?

No.

### DIFF
--- a/cli/gluster-block.c
+++ b/cli/gluster-block.c
@@ -685,11 +685,11 @@ glusterBlockCreate(int argcount, char **options, int json)
     case GB_CLI_CREATE_RBSIZE:
       if (isNumber(options[optind])) {
         sscanf(options[optind++], "%u", &cobj.rb_size);
-        if (cobj.rb_size < 1 || cobj.rb_size > 64) {
-          MSG(stderr, "'ring-buffer' should be in range [1MB - 64MB]");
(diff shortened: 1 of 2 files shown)


cli/gluster-block.c
```diff
@@ -685,11 +685,11 @@ glusterBlockCreate(int argcount, char **options, int json)
     case GB_CLI_CREATE_RBSIZE:
       if (isNumber(options[optind])) {
         sscanf(options[optind++], "%u", &cobj.rb_size);
-        if (cobj.rb_size < 1 || cobj.rb_size > 64) {
-          MSG(stderr, "'ring-buffer' should be in range [1MB - 64MB]");
+        if (cobj.rb_size < 1 || cobj.rb_size > 1024) {
```
pkalever: @lxbsz As per your commit msg the global limitation is 2GB, then why limit to 1G only ?

lxbsz: The global 2GB total size will be shared by all the targets, usually there will be more than 1 target running, so 2GB for each target makes no sense. 

Before we have calculated all the possible cases that could fill up the ringbuffer, 1GB is more than enough.

pkalever:
@lxbsz merged now. Thanks!

PR #254, opened 2019-10-08
genconfig: fix one compile warning

lxbsz:
block_genconfig.c: In function 'block_gen_config_cli_1_svc':
block_genconfig.c:66:43: warning: '%s' directive output may be truncated writing up to 255 bytes into a region of size 239 [-Wformat-truncation=]
   66 |   snprintf(lun_so, 256, "/backstores/user/%s", block);
      |                                           ^~
In file included from /usr/include/stdio.h:867,
                 from ../utils/utils.h:16,
                 from ../utils/common.h:15,
                 from block_common.h:16,
                 from block_genconfig.c:12:
/usr/include/bits/stdio2.h:67:10: note: '__builtin___snprintf_chk' output between 18 and 273 bytes into a destination of size 256
   67 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   68 |        __bos (__s), __fmt, __va_arg_pack ());
      |        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### Does this PR fix issues?

Fixes: #

### DIFF
--- a/rpc/block_genconfig.c
+++ b/rpc/block_genconfig.c
@@ -22,7 +22,11 @@ getTpgObj(char *block, MetaInfo *info, blockGenConfigCli *blk, char *portal, int
   uuid_t uuid;
   char alias[UUID_BUF_SIZE];
   char alias_10[11] = {'\0', };
-  char lun_so[256] = {'\0', };
+  /*
+   * 256 + extra 32 bytes for string "/backstores/user/" to
+   * suppress the truncated warning when compiling.
+   */
+  char lun_so[288] = {'\0', };
 
   struct json_object *tpg_obj = json_object_new_object();
   struct json_object *tpg_attr_obj = json_object_new_object();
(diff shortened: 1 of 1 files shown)


rpc/block_genconfig.c
```diff
@@ -22,7 +22,7 @@ getTpgObj(char *block, MetaInfo *info, blockGenConfigCli *blk, char *portal, int
   uuid_t uuid;
   char alias[UUID_BUF_SIZE];
   char alias_10[11] = {'\0', };
-  char lun_so[256] = {'\0', };
+  char lun_so[384] = {'\0', };
```
pkalever: @lxbsz how did you calculate this ? why not keep it 512 ?

lxbsz: The 256 will be the length of the "char *block", which comes from the d_name, and the 128 will be "/backstores/user/" string.

I am okay with the 512, 384 here just wanted to make sure the array size which hosted in the C stack as small as possible.

pkalever: Okay. If you want to keep the stack small then 128 is also bigger for fixed string here.

strlen("/backstores/user/") = 17

17 + 256, will be good enough. But I will leave it to you, if you want to keep it as it is. In any case making a small comment about the calculation might help the readers.

Thanks!

lxbsz: Yeah, the comment here make sense.
Just add 32 extra bytes instead.
Thanks.

pkalever:
@lxbsz merged now. Thanks!

PR #256, opened 2019-10-30
gluster-blockd: fix import error

pkalever:

$ gluster-blockd
[...]
Parameter loglevel_file is now 'info'.
Parameter max_backup_files is now '100'.
[...]
Parameter auto_save_on_exit is now 'false'.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: cannot import name version


pkalever:
@lxbsz please review.

pkalever:
@lxbsz updated tags and merged now. Thanks!

PR #259, opened 2019-12-10
Fix version typo in configure.ac

bluecmd:
Hopefully self-evident


bluecmd:
Maybe this is wrong? I don't see a version 7.3 available, 7.0 seems to be the latest so I'm not sure. This check is failing for my 7.0 build though, so I assumed this would be correct. Anyway, have a look.

bluecmd:
It was because I was missing libglusterfs-dev - so this might be correct:

```
bluecmd@iscsi1:~/gluster-block$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/glusterfs-api.pc
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: glusterfs-api
Description: GlusterFS API
/* This is the API version, NOT package version */
Version: 7.7.0
```

PR #261, opened 2020-01-03
genconfig: switch to string instead of fixed 255 for the volume

lxbsz:
If we specify more than 2 volumes in one genconfig command line,
the genconfig will truncate the volumes.

### Does this PR fix issues?
Yeah

Fixes: #1787371


pkalever:
Tested and it works as expected.

pkalever:
@lxbsz update the tags, this is ready for merge
Reviewed-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

PR #264, opened 2020-02-19
replace: fix authentication on the new-node

pkalever:
Previously:
Always no authentication set on new node

Fix:
Makesure to update authentication details on the new node

### Notes for the reviewer


pkalever:
@lxbsz could you please paste the targetcli ls output with both auth enable and disable, after replace command is executed.

Thanks!

pkalever:
@lxbsz I think you closed it mistakenly. Reopening it.

PR #265, opened 2020-02-29
create: fix failing with io_timeout not support

lxbsz:
If the configshell version is low, even we do not specify the
io-timeout options, it will fail, but the exepcting should just
skip it.

### Does this PR fix issues?

Fixes: #

### DIFF
--- a/cli/gluster-block.c
+++ b/cli/gluster-block.c
@@ -643,7 +643,6 @@ glusterBlockCreate(int argcount, char **options, int json)
   cobj.json_resp = json;
   cobj.mpath = 1;
   cobj.prealloc = 1;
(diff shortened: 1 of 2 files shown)


pkalever:
@lxbsz can you please elaborate the commit-msg, sorry but its bit unclear for me.

pkalever:
@lxbsz may be an example output of behavior with this patch and without this patch in the commit-msg will help.

lxbsz:
> @lxbsz may be an example output of behavior with this patch and without this patch in the commit-msg will help.

Sure, will update it.

lxbsz:
@pkalever 
Updated the tag.
Thanks.

pkalever:
Merged now, thanks @lxbsz

PR #268, opened 2020-03-31
Vagrant: remove patch reverting in targetcli

pkalever:

targetcli is hanging with
* https://github.com/open-iscsi/targetcli-fb/pull/156

This is fixed now with
* https://github.com/open-iscsi/targetcli-fb/pull/165


pkalever:
Thanks @lxbsz, updated the tags, will merge soon.

PR #272, opened 2020-05-12
daemon: switch to batch mode when available

pkalever:
latest targetcli daemon approach set interactive mode as default to make it
appear similar to cli approach, hence to use previous batch mode we need
some global settings.

Depends on: targetcli#pull/167

### DIFF
--- a/daemon/gluster-blockd.c
+++ b/daemon/gluster-blockd.c
@@ -539,9 +539,17 @@ blockNodeSanityCheck(void)
         return ENOMEM;
       }
     } else {
-      if (GB_ASPRINTF(&global_opts, "%s auto_use_daemon=true", tmp) == -1) {
-        GB_FREE(tmp);
-        return ENOMEM;
+      ret = gbRunner("targetcli set global | grep daemon_use_batch_mode");
+      if (!ret) {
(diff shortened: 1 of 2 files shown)


daemon/gluster-blockd.c
```diff
@@ -539,9 +539,17 @@ blockNodeSanityCheck(void)
         return ENOMEM;
       }
     } else {
-      if (GB_ASPRINTF(&global_opts, "%s auto_use_daemon=true", tmp) == -1) {
-        GB_FREE(tmp);
-        return ENOMEM;
+      ret = gbRunner("targetcli set global | grep daemon_use_batch_mode");
```
lxbsz, line 542: or should s/set/get/ ? though we will get the same result, but it may confusing.

pkalever: @lxbsz 
Maybe get global will have to read the details of each option and display the final output, whereas set global will just display the help. So I thought set will serve the purpose.

Thanks!

lxbsz: Both are okay for me.

pkalever:
@lxbsz updated the tags and merged, thanks!

PR #279, opened 2020-09-10
delete: allow when backend file is absent

pkalever:

To defend a rare case like when the gluster bricks goes down/up
randomly, there is a chance that, unlink of backend file return success
but the writev to metadata file about 'ENTRYDELETE: SUCCESS' failed.

In such cases this patch allow delete to progress.

### Does this PR fix issues?

Fixes: BZ#1847425

### Notes for the reviewer


lxbsz (approved):
LGTM.

pkalever:
Thanks @lxbsz, Merged now

PR #284, opened 2020-12-17
delete: fix subsequent delete failures

pkalever:

On delete, say the block target config on all the nodes is successfully
deleted, but the backend file couldn't get unlink may due to BHV full space.
In this case the retry of delete fails repeatedly.


lxbsz (approved):
I didn't forsee any issue of this change.
LGTM.

